PR #39, opened 2016-08-31
Enable GSSAPI_HEIMDAL for kerberos auth

gitdevmod:


fichtner:
This won't work. We use LibreSSL/OpenSSL from ports and this is incompatible with GSSAPI_BASE. You also need to remove GSSAPI_NONE from the SET statement.

gitdevmod:
ok, which version of GSSAPI is compatible with LibreSSL/OpenSSL from ports ?

fichtner:
Both GSSAPI_HEIMDAL and GSSAPI_MIT, but I am unsure which one we should aim for... I have avoided this subject so far ;(

gitdevmod:
https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/kerberos5.html give some informations about mit heidmal. I used the version from base because I can put password in file and use it with kinit command.

gitdevmod:
From manual only heidmal (base and ports) allow password from file. I really do not know what is different between mit and heidmal.

fichtner:
Merged, thanks! :)

PR #43, opened 2017-01-22
Part 1 of 2 - EAP-RADIUS ipsec/strongswan support

GurliGebis:
This pull request is split across the tools and core repositories, this is the tools part.

To be able to use EAP-RADIUS, we have to enable it, which is what I do here.


fichtner:
looks good to me, thanks!

PR #67, opened 2018-01-16
enable LUA support for net/haproxy-devel

fraenki:
Without LUA support HAProxy will fail to start, rendering the HAProxy plugin unusable.
This is required to be able to switch to HAProxy 1.8 for OPNsense 18.1.

Refs https://github.com/opnsense/plugins/issues/224


fichtner:
Sorry, I missed this. Many thanks!

PR #69, opened 2018-02-16
add POP3 and IMAP to nginx

fabianfrz:
@fichtner should we add mod_security3 also for @mimugmail (opnsense/plugins#507)?

I an not sure if mod_security3 and naxsi conflict since both are WAF extensions.


fichtner:
Merged, thanks. No need to move on mod_security3 just yet.

PR #99, opened 2018-10-17
Add apuledctld support for using on PCEngines APU<x> devices (from Smart-Soft)

alexpro:
To implement as described in opnsense/core#2114


fichtner:
a script-based daemon exists and will be merged soon

PR #119, opened 2019-02-22
Kernel and Device config for Raspberry Pi 2

nekoprog:
@fichtner Hope you can review this pull, see if it works or not.


fichtner:
@nekoprog very nice, thanks!

nekoprog:
Have you flash it on bpi? Does it boots properly? Hope you can share the progress with arm device on forum.

fichtner:
I have it on my list, but don't have a lot of time at the moment. Next step is bootstrapping BPI and seeing if that works, then I'd like to build a proper image so we can avoid the bootstrapping. I'll follow up in the forum with a test image then...

PR #128, opened 2019-03-25
Add RPI 2/3 Firmwares

nekoprog:
Copy required firmwares for RPI 2/3 to boot properly.

Required packages for building RPI images:

For RPI2
`pkg install git qemu-user-static arm-gnueabi-binutils u-boot-rpi2 rpi-firmware`

For RPI3
`pkg install git qemu-user-static aarch64-binutils u-boot-rpi3 rpi-firmware`


build/arm.sh
```diff
@@ -45,7 +45,7 @@ if [ -n "${1}" ]; then
 	ARMSIZE=${1}
 fi
 
-ARMIMG="${IMAGESDIR}/${PRODUCT_RELEASE}-arm-${PRODUCT_ARCH}.img"
+ARMIMG="${IMAGESDIR}/${PRODUCT_RELEASE}-${PRODUCT_TARGET}-${PRODUCT_ARCH}-${PRODUCT_DEVICE}.img"
```
fichtner: "arm" was meant for designation of the image type like all other images do, not the target arch. If we append device that should be good enough. What do you think?

fichtner: or swap PRODUCT_ARCH for PRODUCT_TARGET

fichtner:
We need to add the package hint for rpi-firmware somewhere. Or we could check packages at runtime. Doing something similar for core.git runtime deps: https://github.com/opnsense/core/blob/master/Makefile#L166 (arcane expansion, checks for package installed in other targets...)

build/arm.sh
nekoprog: I think swap PRODUCT_ARCH with PRODUCT_TARGET is the best option. And we should do this with kernel.txz too.

nekoprog:
> We need to add the package hint for rpi-firmware somewhere. Or we could check packages at runtime. Doing something similar for core.git runtime deps: https://github.com/opnsense/core/blob/master/Makefile#L166 (arcane expansion, checks for package installed in other targets...)

Can this check option works if we add some line on device config?

build/arm.sh
fichtner: I agree with the first one. It's also true for the kernel, but in this case we need opnsense-update to be able to have device support, which it currently hasn't. Ideally that is what we wanted to avoid with high hopes for inter-device compatible arm64 images.

device/bpi.conf
```diff
@@ -8,3 +8,4 @@ UBLDR_LOADADDR=0x42000000
 export PRODUCT_KERNEL=SMP-BPI
 export PRODUCT_TARGET=arm
 export PRODUCT_ARCH=armv6
+export WANTS=arm-gnueabi-binutils qemu-user-static u-boot-bananapi
```
fichtner:
```suggestion
export PRODUCT_WANTS="arm-gnueabi-binutils qemu-user-static u-boot-bananapi"
```

not sure about qemu-user-static, think about native builds. but can always fix later

device/rpi2.conf
```diff
@@ -8,3 +8,4 @@ UBLDR_LOADADDR=0x2000000
 export PRODUCT_KERNEL=SMP-RPI2
 export PRODUCT_TARGET=arm
 export PRODUCT_ARCH=armv6
+export WANTS=arm-gnueabi-binutils qemu-user-static u-boot-rpi2 rpi-firmware
```
fichtner:
```suggestion
export PRODUCT_WANTS="arm-gnueabi-binutils qemu-user-static u-boot-rpi2 rpi-firmware"
```

device/rpi3.conf
```diff
@@ -8,3 +8,4 @@ UBLDR_LOADADDR=0x200
 export PRODUCT_KERNEL=SMP-RPI3
 export PRODUCT_TARGET=arm64
 export PRODUCT_ARCH=aarch64
+export WANTS=aarch64-binutils qemu-user-static u-boot-rpi3 rpi-firmware
```
fichtner:
```suggestion
export PRODUCT_WANTS="aarch64-binutils qemu-user-static u-boot-rpi3 rpi-firmware"
```

build/common.sh
```diff
@@ -553,7 +553,7 @@ setup_kernel()
 {
 	echo ">>> Setting up kernel in ${1}"
 
-	tar -C ${1} -xpf ${SETSDIR}/kernel-*-${PRODUCT_ARCH}.txz
+	tar -C ${1} -xpf ${SETSDIR}/kernel-*-${PRODUCT_ARCH}-${PRODUCT_DEVICE}.txz
```
fichtner: can't do this right now without breaking production upgrades

build/kernel.sh
```diff
@@ -31,7 +31,7 @@ SELF=kernel
 
 . ./common.sh
 
-KERNEL_SET=$(find ${SETSDIR} -name "kernel-*-${PRODUCT_ARCH}.txz")
+KERNEL_SET=$(find ${SETSDIR} -name "kernel-*-${PRODUCT_ARCH}-${PRODUCT_DEVICE}.txz")
```
fichtner: same production issue here

nekoprog:
Right, I forgot about other arch build, will fix that with a case later.

build/arm.sh
nekoprog: Update the name with the first one then.

build/common.sh
nekoprog: Can review latest commit?

build/kernel.sh
nekoprog: Can review latest commit?

build/common.sh
```diff
@@ -553,7 +553,14 @@ setup_kernel()
 {
 	echo ">>> Setting up kernel in ${1}"
 
-	tar -C ${1} -xpf ${SETSDIR}/kernel-*-${PRODUCT_ARCH}.txz
+	case ${PRODUCT_TARGET} in
+	arm|arm64)
```
fichtner:
```suggestion
	arm*)
```

fichtner: Hmm, on second thought can we leave this out for a little bit? base needs name updates as well and invoking "tar" twice for the same thing in different paths is never. it also affects rename.sh and clean.sh targets

nekoprog: Right, I also forgot about package extraction. Need to hold these product name append until everything related to it changed too.

fichtner: yes, for now please revert common.sh and kernel.sh -- the other changes to arm are perfectly fine

nekoprog: common and kernel reverted

fichtner:
Merged, thanks a lot! :)

nekoprog:
That's fast! Sure no problem. Will find other way to make the kernel production specific to arm devices.

PR #133, opened 2019-04-24
build arm: fix bananapi.dtb file name

nekoprog:
Need to test bananapi image first from https://drive.google.com/drive/folders/1bAWuECNJwu0LdJOpaIn-2MrTKZ8evGNY. If it boots, this should be safe to merge. Else need to rename the copied bananapi.dtb too.

### DIFF
--- a/build/arm.sh
+++ b/build/arm.sh
@@ -100,7 +100,7 @@ cp -p ${STAGEDIR}/boot/ubldr.bin ${STAGEDIR}/boot/msdos/ubldr.bin
 
 case "${PRODUCT_DEVICE}" in
 bpi)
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks. I'll try to offer a bpi image very soon and help with testing.

mimugmail:
I just check your version on GDrive and a fresh compiled image, both doesn't boot with BPI R1 or R2

nekoprog:
Yep, it's only for bananapi, i don't which board, maybe the first model without any number. Because uboot also have bananapi2 and so on. Maybe @fichtner can show the device because he has one.

fichtner:
next thing I'll do now that 19.1.7 is out...

nekoprog:
If still not boot, it might need to use dd with the included .bin from u-boot-bananapi package. I noticed that all allwinner have a different way to boot into ubldr.

PR #135, opened 2019-04-29
arm: add nanopi-neo and nanopi-neo2

nekoprog:
Placeholder config for NanoPi NEO2 and NanoPi R1 board, will be available to use after Pull Request and Bugzilla is accepted.
~NanoPi NEO2 [PR](https://github.com/freebsd/freebsd-ports/pull/98) and [Bugzilla](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236965)~
NanoPi R1 [Bugzilla](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=237719)

### DIFF
--- a/build/arm.sh
+++ b/build/arm.sh
@@ -106,7 +106,7 @@ aarch64)
 esac
 
 case "${PRODUCT_DEVICE}" in
-bpi)
+bpi-m1)
 	cp -p ${STAGEDIR}/boot/dtb/bananapi.dtb ${STAGEDIR}/boot/msdos/sun7i-a20-bananapi.dtb
 	cp -p /usr/local/share/u-boot/u-boot-bananapi/* ${STAGEDIR}/boot/msdos
 	;;
@@ -129,6 +129,7 @@ orangepi-pc2)
 rpi2)
 	cp -p ${STAGEDIR}/boot/dtb/bcm2836-rpi-2-b.dtb ${STAGEDIR}/boot/msdos/bcm2836-rpi-2-b.dtb
 	cp -p /usr/local/share/u-boot/u-boot-rpi2/* ${STAGEDIR}/boot/msdos
+	cp -p /usr/local/share/rpi-firmware/barmstub8.bin ${STAGEDIR}/boot/msdos
 	cp -p /usr/local/share/rpi-firmware/bcm2709-rpi-2-b.dtb ${STAGEDIR}/boot/msdos
 	cp -p /usr/local/share/rpi-firmware/bootcode.bin ${STAGEDIR}/boot/msdos
 	cp -p /usr/local/share/rpi-firmware/config.txt ${STAGEDIR}/boot/msdos
@@ -139,6 +140,7 @@ rpi2)
 rpi3)
 	#cp -p ${STAGEDIR}/boot/dtb/bcm2837-rpi-3-b.dtb ${STAGEDIR}/boot/msdos/bcm2837-rpi-3-b.dtb
 	cp -p /usr/local/share/u-boot/u-boot-rpi3/* ${STAGEDIR}/boot/msdos
+	cp -p /usr/local/share/rpi-firmware/barmstub8.bin ${STAGEDIR}/boot/msdos
 	cp -p /usr/local/share/rpi-firmware/bcm2710-rpi-3-b.dtb ${STAGEDIR}/boot/msdos
 	cp -p /usr/local/share/rpi-firmware/bcm2710-rpi-3-b-plus.dtb ${STAGEDIR}/boot/msdos
 	cp -p /usr/local/share/rpi-firmware/bootcode.bin ${STAGEDIR}/boot/msdos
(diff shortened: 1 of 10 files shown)


build/arm.sh
```diff
@@ -93,16 +93,31 @@ EOF
...
 case "${PRODUCT_DEVICE}" in
 bpi)
 	cp -p ${STAGEDIR}/boot/dtb/bananapi.dtb ${STAGEDIR}/boot/msdos/sun7i-a20-bananapi.dtb
 	cp -p /usr/local/share/u-boot/u-boot-bananapi/* ${STAGEDIR}/boot/msdos
 	;;
+nanopi-neo2
+	cp -p ${STAGEDIR}/boot/dtb/sun50i-h5-nanopi-neo2.dtb ${STAGEDIR}/boot/msdos/sun50i-h5-nanopi-neo2.dtb
+	cp -p /usr/local/share/u-boot/u-boot-nanopi-neo2/* ${STAGEDIR}/boot/msdos
+	;;
+nanopi-r1
```
fichtner: can you rename this one to nanopi-neo like in u-boot?

nekoprog: Sure, hope that NEO image will be compatible with NanoPi R1.

nekoprog:
Wait, need to change other files to neo too.

fichtner:
is r1 not the same as "neo" ? I was confused by the u-boot file set... if it needs a different kernel config then neo then my suggestion is not the best one.

nekoprog:
It's not same, R1 got many hardwares, but they uses same CPU. I've rename all R1 to NEO, it should be safe now. NEO got shitty 512mb RAM, while R1 got less shitty 1GB RAM. I have to wait for u-boot to update their devconfig first so that freebsd can make new port for u-boot-nanopi-r1.

nekoprog:
Can wait a little more before merging? I want to rename bananapi to avoid future confusion.

fichtner:
Ok, please do what you think is best. I'm not up to speed on all the devs out there for each vendor and let me know when you want to have it merged :)

nekoprog:
@fichtner alright, done for now.

device/bpi-m1.conf
```diff
@@ -1,11 +1,12 @@
-# http://www.banana-pi.org/
+# http://www.banana-pi.org/m1.html
+# http://www.banana-pi.org/m1plus.html
```
fichtner: One question.. I have the R1 and isn't it the same as M1 from boot perspective?

nekoprog: from SMP kernel it's same because they use allwinner kernel conf. But I'm not sure from device tree blob (dtb) copied from u-boot. but report from @mimugmail he said that current bpi image won't boot. maybe current fix for allwinner I add on arm.sh will do the trick?

fichtner: let me take care of this... @mimugmail gave me a m1 and r1 for that particular reason a while back :)

nekoprog: nice. hope that this image will boot. R1 looks like a nice board, got many wan/lan options.

fichtner: maybe some day r64 works, that one looks very nice.. it has PoE as well

mimugmail:
I thought a send you R1 and R2? Way too long ago :)

Am Fr., 3. Mai 2019 um 09:56 Uhr schrieb Neko Prog <notifications@github.com

device/bpi-m1.conf
nekoprog: i'll take a look at it asap. but why all these boards only use 1gb ram? it got many potentials if have at least 2gb. maybe can hack the hardware to upgrade it's ram?

build/arm.sh
```diff
@@ -93,16 +103,31 @@ EOF
...
-cp -p ${STAGEDIR}/boot/ubldr ${STAGEDIR}/boot/msdos/ubldr
-cp -p ${STAGEDIR}/boot/ubldr.bin ${STAGEDIR}/boot/msdos/ubldr.bin
+
+case "${PRODUCT_ARCH}" in
+armv6)
+	cp -p ${STAGEDIR}/boot/ubldr ${STAGEDIR}/boot/msdos/ubldr
+	cp -p ${STAGEDIR}/boot/ubldr.bin ${STAGEDIR}/boot/msdos/ubldr.bin
+	;;
+aarch64)
+	cp -p ${STAGEDIR}/boot/loader.efi ${STAGEDIR}/boot/msdos/bootaa64.efi
```
nekoprog: @fichtner will you add this part of EFI loader into current arm.sh? taken from https://gitlab.com/HardenedBSD/hardenedBSD/blob/hardened/current/master/release/arm64/PINE64.conf#L37. This PR need to be closed because got a big conflict on arm.sh. Maybe I'll revert changes on everything except for nanopi device config for it to be merged.

fichtner: let's please close this and start on separate pieces again. so far RPI3 is not officially supported on 11.2 which makes me doubt we should try it now and I want the arm code to be more structured in their device files (u-boot code in there is better for readability).

nekoprog:
closing, will open new PR later if needed.

PR #139, opened 2019-05-07
device: in case BANANAPI img won't boot

nekoprog:
@fichtner 
Copying dtb files to msdos directory instead of dtb directory.


fichtner:
This doesn't work any better. I get the u-boot boot and preliminary loader, but when the kernel is going up the HDM dies with a bit of static and then it's off. The FreeBSD BANANAPI image has the same issue. Will try to see how Linux does it...

nekoprog:
What a weird way to boot. I blame sunxi for it's complexity.

fichtner:
it works, but dtb is likely flawed, FreeBSD uses its own

PR #141, opened 2019-05-09
keep HAProxy 1.8.x

fraenki:
The FreeBSD port net/haproxy was recently switched to the 1.9 release series of HAProxy. As stated in https://github.com/opnsense/plugins/issues/1089 I'd rather stay on 1.8.x until 2.0 was released and stabilized.

HAProxy 1.8 moved to a new port:
https://www.freshports.org/net/haproxy18

### DIFF
--- a/config/19.1/ports.conf
+++ b/config/19.1/ports.conf
@@ -109,6 +109,7 @@ net/frr5					arm,arm64
 net/frr6					arm,arm64
 #net/frr7					arm,arm64
 net/haproxy					arm,arm64
+net/haproxy18					arm,arm64
 net/hostapd
 net/igmpproxy
 net/isc-dhcp44-relay


fichtner:
I haven't replaced the port version, we can keep "haproxy" and don't need to side-track unless you want to add this as the separate port and switch later if 1.8 updates keep happening

fraenki:
@fichtner The new port net/haproxy18 is fully maintained and receives updates, for example the new 1.8.20 release which contains many interesting fixes. I think it would be better mid-term to switch to this new port, because I'm pretty sure it will take some time for HAProxy 2.0 to stabilize.

fichtner:
then let's have both ports in the config

fraenki:
I've updated the PR accordingly, just in case you'd like to merge it.

fichtner:
merged, thanks ❤️

Issue #115, opened 2019-02-14
obsolete file generation issue


evbevz:
https://github.com/opnsense/tools/blob/0eb37037bbe29ffb864c58e794c66baad987aceb/config/19.1/plist.base.amd64#L476

Which collate you use when sorting base plist during base obsolete file generation?
In my case en_US.UTF-8:
/lib/libz.so.6
/lib/libzfs_core.so.2
/lib/libzfs.so.2

In your pkg:
/lib/libz.so.6
/lib/libzfs.so.2
/lib/libzfs_core.so.2
 
As result I have /lib/libzfs.so.2 in my base.obsolete file. It is problem for me.

fichtner:
Hmm, I haven't set up anything that would change the default sorting behaviour... whatever sort(1) defaults to then...

evbevz:
Maybe some special locale is set on build machine?

evbevz:
https://github.com/opnsense/tools/blob/8d5f94880df51d54fed3dd6c7c47b227bb88db84/config/18.7/plist.base.amd64#L512
Here is other sort order.

fichtner:
Not that I'm aware of. I cannot reproduce.

fichtner:
"." is 46, "_" is 95 in ASCII so the sort order seems correct with what we have...

PR #146, opened 2019-07-05
obsolete: fix build machines collate order differences

evbevz:
Last times, if I build base, I get broken obsolete file in result. Because some files has other sort order in generated sorted file list then obsolete file from repository. As result, diff generates '+' and '-' records for this files, as they are moved in sort order. And then filters out '+' records and generates extra records in obsolete file. And during system update, those files deleted and system becomes broken.

### DIFF
--- a/build/base.sh
+++ b/build/base.sh
@@ -88,12 +88,13 @@ echo -n ">>> Generating obsolete file list... "
 
 : > ${STAGEDIR}/setdiff.old
 if [ -f ${CONFIGDIR}/plist.base.${PRODUCT_ARCH} ]; then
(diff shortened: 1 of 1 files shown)


evbevz:
https://github.com/opnsense/tools/issues/115

fichtner:
Instead of stuffing and reshuffling, maybe we should check the root cause. My clean build env says:

```
# locale
LANG=
LC_CTYPE="C"
LC_COLLATE="C"
LC_TIME="C"
LC_NUMERIC="C"
LC_MONETARY="C"
LC_MESSAGES="C"
LC_ALL=
```

Yours is probably set up different so either we need to enforce a clean "C" type or you need to...

PR #147, opened 2019-07-07
add net/google-cloud-sdk

fraenki:
required for https://github.com/opnsense/plugins/pull/1392


fichtner:
I think we can go with this, but this is for config/19.7 as config/19.1 will be dropped soon enough and will not yield any new packages anymore

fraenki:
config/19.7 is fine for me :+1: (Ad requested other changes that will take some time to implement)

fichtner:
merged, thanks!

PR #150, opened 2019-08-08
os-sunnyvalley: we can enable the build for LibreSSL

muratbalaban43:
we can enable the build for LibreSSL since os-sensei does not specifically depend on OpenSSL/LibreSSL for now. The same holds true for elasticsearch5.

### DIFF
--- a/config/19.7/plugins.conf
+++ b/config/19.7/plugins.conf
@@ -62,7 +62,7 @@ sysutils/nut					arm,arm64
 sysutils/smart					arm,arm64
 sysutils/vmware					arm,arm64
 sysutils/xen					arm,arm64
-vendor/sunnyvalley				arm,arm64,i386,LibreSSL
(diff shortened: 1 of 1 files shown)


fichtner:
This is a ticking time bomb if you don’t maintain a separate LibreSSL repository like we do.

muratbalaban43:
@fichtner , I agree. 

For now, since we don't have a dynamic dependency on any ssl library, we're fine. 

When we introduce TLS, and we depend on an ssl library, we'll make sure we have seperate builds for both, thus seperate repositories.

fichtner:
Dynamic linking on dependencies that eventually depend on an SSL library can lead to segfaults if both libraries are being used somewhere down this dynamic link chain. The effects are hard to trace but relatively easy to circumvent: either by restricting to OpenSSL or building both crypto flavours. There isn’t much leeway here.

PR #152, opened 2019-08-13
config: Enable SNMP support for FRR

devmast:
* Added SNMP make option for FRR 5 and 6
* See: http://docs.frrouting.org/en/latest/snmp.html


fichtner:
@devmast @mimugmail how are we going to use this?

devmast:
@fichtner @mimugmail This will allow the "new" net-snmp plugin to expose routing metrics from frr for remote monitoring. Ideally, we would also need to:
*  Add the `master agentx` option to the net-snmp template here: https://github.com/opnsense/plugins/blob/master/net-mgmt/net-snmp/src/opnsense/service/templates/OPNsense/Netsnmp/snmpd.conf
* Add the `agentx` option to the templates here: https://github.com/opnsense/plugins/tree/master/net/frr/src/opnsense/service/templates/OPNsense/Quagga

I have not scoped out the changes beyond that, however just adding the frr snmp build option is a good first step.

mimugmail:
There's already an open issue regarding agentx, for me it's ok :)

fichtner:
Ok, great. Thanks for the context. Merged!

devmast:
Thanks @fichtner @mimugmail 

Is the issue: opnsense/plugins#1200 ? I can have a go at it unless this is something you all want to handle.

mimugmail:
You can take it of course :)

PR #155, opened 2019-09-26
device: several arm device configs for 20.1

nekoprog:
Only to be used with OPNsense 20.1 (FreeBSD 12.1 http://ftp.freebsd.org/pub/FreeBSD/releases/ISO-IMAGES/12.1/)
Product kernel is expected to be GENERIC SMP, no need for separate SMP-PRODUCT_DEVICE.


nekoprog:
Reference: [FreeBSD SVN RPI3.conf](https://svnweb.freebsd.org/base/stable/12/release/arm64/RPI3.conf?view=markup)

fichtner:
I don't feel comfortable adding these just yet.... we still only have a FreeBSD releng/12.1 tree and HBSD as well as all our required patches are missing.

fichtner:
Ok to keep this open, I just wanted to say we aren't ready for merge yet :)

nekoprog:
Thanks a lot, no problem, because I can keep track from here easier than to look from my fork.

nekoprog:
@fichtner I'm thinking about changing all arm to use GENERIC SMP, it will build all kernel for all boards and tools.git will copy them all into /boot/dtb, only few megabytes to include them all. So no separate  SMPs for different arm devices.

fichtner:
@nekoprog yes please that sounds nice

nekoprog:
@fichtner generic kernel for arm 32 and 64 bit can be built successfully using DEVICE=ARM32 or ARM64. What's next is to fix the problem with cross build. I got a fatal error on both xtools and packages. Build steps for base, kernel and arm can run without a problem.

build/xtools.sh
```diff
@@ -56,6 +56,8 @@ MAKE_ARGS="TARGET_ARCH=${PRODUCT_ARCH} TARGET=${PRODUCT_TARGET}"
 MAKE_ARGS="${MAKE_ARGS} SRCCONF=${CONFIGDIR}/src.conf __MAKE_CONF="
 
 ${ENV_FILTER} make -C${SRCDIR} -j${CPUS} native-xtools ${MAKE_ARGS} NO_CLEAN=yes
+# XXX Temporary fix for cross build
+${ENV_FILTER} make -C${SRCDIR} -j${CPUS} native-xtools-install ${MAKE_ARGS} NO_CLEAN=yes
```
fichtner: looks like the proper fix since FBSD 12 introduced the new target and split the install stage from the old, maybe we can get away with changing native-xtools to native-xtools-install in the line above. Unfortunately that like is not backwards compatible so we have to be careful not breaking 11.x builds

nekoprog: Actually this doesn't fix anything because nxb-bin was build on root folder (/nxb-bin) and not the actual OBJDIR. But for the sake of successful build, this is just a temporary fix. I found that many variables introduced by opnsense/tools doesn't retain it's data. And yes this is not backward compatible. only for 12.1.

fichtner:
Can you explain how data is not retained in variables? We do talk about fundamental changes in src.git here, not issues in tools.git because building works fine for 11.

build/xtools.sh
fichtner: well, that is because on top of native-xtools-install they introduced DESTDIR so if it is empty everything expands to "/" root dir instead of the actual stage dir. Not sure about the other thing, but I'm going to check it out as soon as we have hardenedbsd patches in the tree again.

fichtner: 20171028:
        The native-xtools target no longer installs the files by default to the 
        OBJDIR.  Use the native-xtools-install target with a DESTDIR to install 
        to ${DESTDIR}/${NXTP} where NXTP defaults to /nxb-bin.

nekoprog: So that's the reason why some of these got empty string. Check out base and kernel too, I've marked them on my fork with `xxx temporary fix`. Maybe same thing happens on setup_stage().

And my current issue is that when running `make packages`, it complains about these:
```
make: "/etc/make.conf" line 0: 1 open conditional
make: Fatal errors encountered -- cannot continue
[: make:: unexpected operator
```

fichtner: should be 80e28f5e8dc53e

nekoprog: Thanks, I'll test it now. Will let you know the results later.

config/20.1/make.conf
```diff
@@ -30,11 +30,12 @@ OPTIONS_UNSET=		DOCS EXAMPLES GCC GSSAPI_BASE NLS OPENGL WAYLAND X11
 DEFAULT_VERSIONS=	python=${PRODUCT_PYTHON3:C/^./&./}
 DEFAULT_VERSIONS+=	python3=${PRODUCT_PYTHON3:C/^./&./}
 DEFAULT_VERSIONS+=	ruby=${PRODUCT_RUBY:C/^./&./}
+# XXX need to add if product is arm/arm64, go to openssl 1.0.2
 .if ${PRODUCT_FLAVOUR:tl} != "libressl"
```
nekoprog: @fitchner I need help with this, I'm not familiar with makeconf to be able to make it work. Tried with `.if ${PRODUCT_FLAVOUR:tl} != "libressl" && ${PRODUCT_TARGET} != "arm" && ${PRODUCT_TARGET} != "arm64"` but returns malformed conditional.

fichtner: just set FLAVOUR=LibreSSL when building

nekoprog: Noted. Will try that on next build.

nekoprog:
@fichtner I'm done with ARM for 20.1 for now, I hand it over to you if you want to merge it or not. Want to learn more about ARM and UEFI next, will try to make one config for all ARM devices if I can, no need to rely on uboot.

DarkSunOne:
@nekoprog not all arm devices support efi, i cant imagine, creating an 'universal' arm image.
Also there are still a lot custom kernel parameters for some devices (specially the sg3100 with the dtb stuff)

Also your RPI3 config wont work, cause there is no "boot.scr" anymore (theres only a efi stub)
Further just see the differences in the raspberry config files, i think its not that ease, to include all of them into one config :-(

But at least it should be possible, to use the same sets for all these builds, so only the 'make arm' needs to be individual for all that devices.

nekoprog:
I'm not looking back to armv7, looking forward to aarch64 and efi. RPI variant not a good example for this.

PR #160, opened 2019-11-06
openssl: disabled 128 bit comiler support / miniupnpd: disabled check_port_inuse

DarkSunOne:
openssl: 128bit compiler support is not needed anymore (and its even not supported anymore)

miniupnpd: [https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236173](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236173)


DarkSunOne:
Oh crap, just noticed that "net_miniupnpd_UNSET" isn't needed, cause the option is enabled on line before >.<

PR #161, opened 2019-11-06
openssl: disabled 128bit compiler / miniupnpd: disabled check_portinuse

DarkSunOne:
openssl: 128bit compiler support is not needed anymore (and its even not supported anymore)

miniupnpd: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236173


config/20.1/make.conf
```diff
@@ -76,7 +76,7 @@ security_fwknop_UNSET=		IPFW
 security_openconnect_SET=	GNUTLS
 security_openconnect_UNSET=	GSSAPI
 security_openssh-portable_UNSET=HPN
-security_openssl_UNSET=		SSL2 SSL3
+security_openssl_UNSET=		SSL2 SSL3 enable-ec_nistp_64_gcc_128
```
fichtner:
```suggestion
security_openssl_UNSET=		EC SSL2 SSL3
```

Shouldn't this be the proper thing? We unset the option, not the build details :)

config/20.1/make.conf
```diff
@@ -63,7 +63,7 @@ net_freeradius3_UNSET=		HEIMDAL
 net_frr6_SET=			SNMP
 net_frr7_SET=			SNMP
 net_haproxy_SET=		LUA
-net_miniupnpd_SET=		CHECK_PORTINUSE PF_FILTER_RULES
+net_miniupnpd_SET=		PF_FILTER_RULES
```
fichtner:
```suggestion
net_miniupnpd_SET=		CHECK_PORTINUSE  PF_FILTER_RULES
```

Here this is more complicated, it's broken in FreeBSD somewhat, 12 ok, 11 not ok: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=233963

we just flipped the logic but ideally FreeBSD must be fixed to enable both to work again

sorry to make this so complicated, but we need the most flexible fix so that it doesn't break left and right and enables us to not worry about it anymore

config/20.1/make.conf
DarkSunOne: Well, i still have to learn a lot here ;-)

config/20.1/make.conf
DarkSunOne: Oh crap, that's so screwed up :-(
But at least there is some progress.

I completely agree to stay as near as possible to the FreeBSD source (and the options),
but for now, it's just not possible to build with that bug, and there is a working workaround.

For now i think that this is the only one, which completely breaks the build procedure, so why not temporary apply that workaround, until its fixed by FreeBSD?

fichtner: because there's nothing more persistent than a temporary workaround

fichtner: (I asked for commit in the freebsd ticket, the patch was ready but forgotten it seems)

DarkSunOne: So well, i definitely have to agree with that ^^

So lets keep everything in place, and hope that they will commit it asap.

nekoprog: Almost half a year later. Need to ping FreeBSD to apply the patch soon.

config/20.1/make.conf
```diff
@@ -63,7 +63,7 @@ net_freeradius3_UNSET=		HEIMDAL
 net_frr6_SET=			SNMP
 net_frr7_SET=			SNMP
 net_haproxy_SET=		LUA
-net_miniupnpd_SET=		CHECK_PORTINUSE PF_FILTER_RULES
+net_miniupnpd_SET=		CHECK_PORTINUSE  PF_FILTER_RULES
```
fichtner:
```suggestion
net_miniupnpd_SET=		CHECK_PORTINUSE PF_FILTER_RULES
```

whoopsie, shouldn't touch this line... stray whitespace

fichtner:
Merged, thanks!

PR #169, opened 2020-01-28
security/acme.sh: set EXAMPLES to install dnsapi files

fraenki:
fixes https://github.com/opnsense/plugins/issues/1678


fraenki:
@fichtner Is there any chance to get this in for 20.1.0? Otherwise the ACME plugin remains nonfunctional for many users.

fichtner:
Hiding functional files under EXAMPLES flag which we or pfSense never used sounds like a bad idea. I'll revert in ports but this will take considerable time to fix in 19.7.10_1 and 20.1 that I don't have.  Bye bye QA time for 20.1 ;)

PR #181, opened 2020-06-05
Fixed object paths for cross compiling

DarkSunOne:
As complained in [https://wiki.freebsd.org/A_Brief_Guide_To_Cross_Compiling_FreeBSD](https://wiki.freebsd.org/A_Brief_Guide_To_Cross_Compiling_FreeBSD), where $TARGET != $TARGET_ARCH, the object dir is "$MAKEOBJDIRPREFIX/$TARGET.$TARGET_ARCH"

So in this case -> /usr/obj/$TARGET.$TARGET_ARCH

further I enabled armv7

### DIFF
--- a/config/20.7/make.conf
+++ b/config/20.7/make.conf
@@ -63,6 +63,7 @@ net_freeradius3_SET=		LDAP MITKRB_PORT MYSQL SQLITE3
 net_freeradius3_UNSET=		HEIMDAL
 net_frr7_SET=			SNMP
 net_haproxy_SET=		LUA
+net_haproxy20_SET=		LUA
 net_miniupnpd_SET=		CHECK_PORTINUSE PF_FILTER_RULES
 net_openldap24-server_SET=	MEMBEROF REFINT SASL
 net_vnstat_UNSET=		GUI


build/base.sh
```diff
@@ -58,7 +58,11 @@ BASE_OBJDIR="$(make -C${SRCDIR}/release -V .OBJDIR)"
 setup_stage "${BASE_OBJDIR}/${BASE_DISTDIR}"
 
 # remove older object archives, too
-BASE_OBJ=$(make -C${SRCDIR}/release -V .OBJDIR)/base.txz
+if [ ${PRODUCT_HOST} != ${PRODUCT_ARCH} ]; then
```
fichtner, line 61: hmm, ok, not very portable but ok... I'm simply wondering if this check is actually correct and if this breaks compat for 11.2 cross-building which was supposed to be working fine.

DarkSunOne: Yes, this will definitely break cross compiling on BSD 11.2 (didn't thought about this).

So whats the solution for this?
Include an exception for 11.2, or wait until 11.2 (OPNsense 19.7) is obsolete?

DarkSunOne: Well .... screwed up my git again, let me fix that

build/base.sh
```diff
@@ -58,7 +58,11 @@ BASE_OBJDIR="$(make -C${SRCDIR}/release -V .OBJDIR)"
 setup_stage "${BASE_OBJDIR}/${BASE_DISTDIR}"
 
 # remove older object archives, too
-BASE_OBJ=$(make -C${SRCDIR}/release -V .OBJDIR)/base.txz
+if [ ${PRODUCT_HOST} != ${PRODUCT_ARCH} -a ${SRCREVISION} != 11.2 ]; then
+	BASE_OBJ=$(make -C/usr/obj${SRCDIR}/${PRODUCT_TARGET}.${PRODUCT_ARCH}/release -V .OBJDIR)/base.txz
+else
+	BASE_OBJ=$(make -C${SRCDIR}/release -V .OBJDIR)/base.txz
```
fichtner, line 64: I wonder if passing ${MAKE_ARGS} here is enough to fix this? Also we don't need the 11.2 workaround anymore

DarkSunOne: Finally :))
i just started working again on arm images from stretch.
i will update this issue as soon i got there.

PR #191, opened 2020-08-02
keep HAProxy 2.0, upstream port was renamed

fraenki:
The upstream port was renamed from `net/haproxy` to `net/haproxy20`. Now the old port is HAProxy 2.2, but we should keep HAProxy 2.0 for the time being.

### DIFF
--- a/config/20.7/ports.conf
+++ b/config/20.7/ports.conf
@@ -106,7 +106,7 @@ net/freeradius3					arm,arm64
 net/freevrrpd					arm,arm64
 net/frr7					arm,arm64
 net/google-cloud-sdk				arm,arm64
-net/haproxy					arm,arm64
+net/haproxy20					arm,arm64
 net/hostapd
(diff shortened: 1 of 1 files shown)


fichtner:
shuffling this around is pretty useless if we end up using "haproxy" package again eventually. I'd rather make sure to sync haproxy20 to haproxy or at least enable both packages to coexist...

fraenki:
Yes, at some point in the future we'll be using "haproxy" again, but I think this is several months away.

fraenki:
@fichtner I thought this is the best solution because it would keep maintenance low, but do what fits best :)

fichtner:
I just see intermittent breakage moving the package, then moving plugins then moving package then moving back. Plus we don't have a way to use the dev version.

How about adding haproxy20 package. Then switch one or both plugins over after 20.7.1?

fraenki:
Sure :+1:

PR #198, opened 2020-09-17
HBSD: Disable SHLIBRANDOM by default

lattera:
It shouldn't be enabled by default, anyways. Something's gone screwy.
More investigation needed.

Signed-off-by:	Shawn Webb <shawn.webb@hardenedbsd.org>


fichtner:
We only need https://github.com/opnsense/src/commit/dc1b6c6278 since both apply to build time. Thanks!

PR #201, opened 2020-10-30
Remove unset GUI from vnstat build config.

ladar:
In theory this will cause `vnstati` to get built. The catch is that now the `gd` package will become a required dependency for `vnstat` ...

This is the first step towards addressing https://github.com/opnsense/ports/issues/103


fichtner:
I have to decline this for the GUI /graphics based ports we do not require and like to would point to other efforts to improve statistic extraction and visualisation here instead https://github.com/opnsense/plugins/pull/2073#issuecomment-716192677

PR #203, opened 2020-11-17
Working ARM configs for 20.7

nekoprog:
Updated several files to match with OPNsense 20.7 and FreeBSD 12.1
requirements


nekoprog:
@fichtner this fixes are ready to merge. Successfully build all sets files.

fichtner:
@nekoprog sorry to have kept you wanting... can you move your changes to 21.1 since 20.7 won't be supported beyond end of January?

nekoprog:
Sure, will let you know when it's done

bevobird:
Is it normal that it takes like 4-5 hours, and he is still not finished with the build ?

fichtner:
Sure, with all the community requests for binary packages and plugins you are looking at what we build and maintain every day. :)

bevobird:
Can you tell me how long it can take ? so like 8 hours or like this ?

fichtner:
It depends on a lot of factors. Our nightly build on amd64 fast hardware takes about 6 hours from scratch https://nightly.opnsense.org/20.7/amd64/logs 

If we walk about arm here IDK whether it's cross-compile (not as fast as amd64 native but still ok) or native (can take more than 24 hours)

nekoprog:
> Can you tell me how long it can take ? so like 8 hours or like this ?

Very long if you have under 6 core cpus, over 24 hours, and take even longer on make packages.

nekoprog:
KIV, wait for 21.7

nekoprog:
hi, sorry for late reply, lately i have a tight schedule on my farm, so don't have much spare time to resume this work before 21.7 release. but i will try to come here often and guide those who wants to resume my work.

fichtner:
@nekoprog not a problem, thanks for all your work and see you soon :)

PR #213, opened 2021-01-30
Add beats7 to ports.conf

itn3rd77:
As Elasticsearch 8 is on it's way and beats6 isn't compatible with Elastic Common Schema (ECS) I would like to finally add beats7 before the Elasticsearch 8 release.


fichtner:
Had to remove some others, see https://github.com/opnsense/tools/commit/1afa5a39f1

PR #220, opened 2021-03-21
add new port/plugin: emulators/qemu-guest-agent

fraenki:
required for https://github.com/opnsense/plugins/pull/2293


config/21.1/make.conf
```diff
@@ -53,6 +53,7 @@ databases_rrdtool_UNSET=	DEJAVU GRAPH
 devel_git_UNSET=		GITWEB SEND_EMAIL SVN SUBTREE
 dns_getdns_SET=			LIBEV LIBEVENT LIBUV
 dns_unbound_SET=		PYTHON
+emulators_qemu-guest-agent_SET=		PIE RELRO SAFESTACK CFI RETPOLINE SLH
```
fichtner, line 56: We don't set HBSD flags from here.

```suggestion
```

fichtner:
Builds in 40 seconds so this is more than good. Thanks!

PR #221, opened 2021-03-29
adjust make.conf sets of removed ports

mimugmail:


config/20.7/make.conf
```diff
@@ -94,9 +94,6 @@ sysutils_flashrom_UNSET=	FTDI
 sysutils_msktutil_SET=		${PRODUCT_GSSAPI}
 www_lighttpd_UNSET=		LUA
 www_nginx_SET=			BROTLI NAXSI MAIL_IMAP MAIL_POP3 NJS VTS
-www_node_SET=			BUNDLED_SSL
```
fichtner, line 97: 20.7 still has www/node

mimugmail: indeed .. sry :)

fichtner:
Merged, thanks!

PR #222, opened 2021-04-05
Working aarch64 configs for 21.1

yrzr:
Based on the work of @nekoprog.

I have successfully built aarch64 images for RPi3 and virtual machine images based on the scripts.

Moreover, I am using an RPi4 to build the images. Cross-building doesn't work at least on my machine with the kernel building stage since 21.1.


nekoprog:
hi, thanks for the work, i can't review all the changes you've done, but i will try to help if there's a problem

config/21.1/aux.conf
```diff
@@ -1,3 +1,3 @@
 #ORIGIN						IGNORE
-lang/go						arm,arm64
+lang/go
```
fichtner, line 2: which go applications are not properly marked in ports.conf ? This doesn't change a lot since go will still be built in a ports run of needed and historically we tried to avoid go in any architecture that is not amd64. I'm not opposed either way, just that not a lot of go ports are enabled for arm64

config/21.1/plugins.conf
```diff
@@ -1,78 +1,78 @@
 #ORIGIN						IGNORE
-benchmarks/iperf				arm,arm64
```
fichtner, line 2: changes are a bit intrusive here... I don't see a clear agenda?

fichtner:
@yrzr thanks for this... in general yes this should go in as previously discussed with @nekoprog but we need to break it down a little into parts required for building and tweaking shared settings like aux.conf/ports.conf

device/CLEARFOG.conf
```diff
@@ -0,0 +1,23 @@
+# https://www.solid-run.com/marvell-armada-family/clearfog/
+
+export MAKE_ARGS_DEV="
+CROSS_BINUTILS_PREFIX=/usr/local/arm-none-eabi/bin
```
nekoprog, line 4: this should use `CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd${PRODUCT_ABI}/bin`

because arm-none-eabi will cause many problem when compiling kernel, especially with cloudabi and [armelf_fbsd](https://github.com/opnsense/src/blob/master/sys/conf/kern.mk#L280)
if you found any workaround instead of manually editing kern.mk to make it useable with none-eabi, it would be great.

this applies to all armv7h

device/RPI2.conf
```diff
@@ -1,13 +1,14 @@
 # https://www.raspberrypi.org/products/raspberry-pi-2-model-b/
 
 export MAKE_ARGS_DEV="
-CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd11.2/bin
```
nekoprog, line 4: this should use `CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd${PRODUCT_ABI}/bin`

because arm-none-eabi will cause many problem when compiling kernel, especially with cloudabi and [armelf_fbsd](https://github.com/opnsense/src/blob/master/sys/conf/kern.mk#L280)
if you found any workaround instead of manually editing kern.mk to make it useable with none-eabi, it would be great.

this applies to all armv7h

device/RPI3.conf
```diff
@@ -0,0 +1,32 @@
+# https://www.raspberrypi.org/products/raspberry-pi-3-model-b/
+# https://www.raspberrypi.org/products/raspberry-pi-3-model-b-plus/
+
+export MAKE_ARGS_DEV="
+CROSS_BINUTILS_PREFIX=/usr/local/aarch64-unknown-freebsd12.1/bin
```
nekoprog, line 5: future proof this config using ${PRODUCT_ABI}
`CROSS_BINUTILS_PREFIX=/usr/local/aarch64-unknown-freebsd${PRODUCT_ABI}/bin`

nekoprog (commented):
some insight

config/21.1/aux.conf
yrzr: Yes, you are right, go will still be built in the ports building stage when needed.

config/21.1/plugins.conf
yrzr: Here I just want to enable as many packages as possible for the arm. Most of the packages here are already supported in FreeBSD 12.1 ports, and I have also successfully built them.

yrzr:
@fichtner maybe I shall break down the changes into several parts so that it will be easier for you to review.

device/RPI2.conf
yrzr: Done, thanks

yrzr:
@fichtner let's postpone the changes of ports.conf/plugins.conf for now. I have re-committed, please take a look ;)

build/common.sh
```diff
@@ -244,7 +256,9 @@ fi
 . ${DEVICEDIR}/${PRODUCT_DEVICE_REAL}.conf
 
 # reload the kernel according to device specifications
-export PRODUCT_KERNEL="${PRODUCT_KERNEL}${PRODUCT_DEVICE+"-${PRODUCT_DEVICE}"}"
+if [ ${PRODUCT_ARCH} != armv6 -a ${PRODUCT_ARCH} != armv7 -a ${PRODUCT_ARCH} != aarch64 ]; then
```
fichtner, line 259: actually that was supposed to be an ARM thing, but now it's applied to anything not ARM

This would be more appropriate with a device config and using the following:

https://github.com/opnsense/tools/blob/17fa565a1589d029169504b3414d9aea4532b193/device/A10.conf#L12

build/common.sh
```diff
@@ -66,7 +71,11 @@ while getopts ${OPTS} OPT; do
 		export PLUGINSBRANCH=${OPTARG}
 		;;
 	F)
-		export PRODUCT_KERNEL=${OPTARG}
+		if [ ${PRODUCT_ARCH} == aarch64 ]; then
+				export PRODUCT_KERNEL=${OPTARG}-ARM
```
fichtner, line 75: also better done from device config file

build/common.sh
```diff
@@ -229,6 +238,9 @@ SHELL=${SHELL} BLOCKSIZE=${BLOCKSIZE} MAIL=${MAIL} PATH=${PATH} \
 TERM=${TERM} HOSTTYPE=${HOSTTYPE} VENDOR=${VENDOR} OSTYPE=${OSTYPE} \
 MACHTYPE=${MACHTYPE} PWD=${PWD} GROUP=${GROUP} HOST=${HOST} \
 EDITOR=${EDITOR} PAGER=${PAGER} ABI_FILE=${ABI_FILE}"
+if [ ${PRODUCT_ARCH} == armv6 ] || [ ${PRODUCT_ARCH} == armv7 ] || [ ${PRODUCT_ARCH} == aarch64 ]; then
+       export ENV_FILTER="${ENV_FILTER} ALLOW_UNSUPPORTED_SYSTEM=yes"
```
fichtner, line 242: it's in the make.conf. Why do we need it here?

device/RPI2.conf
fichtner: Actually 11.2 corresponds to SRCABI, not PRODUCT_ABI

device/RPI3.conf
fichtner:
```suggestion
CROSS_BINUTILS_PREFIX=/usr/local/aarch64-unknown-freebsd${SRCABI}/bin
```

build/common.sh
fichtner: That should be in RPI3.conf it seems

build/arm.sh
```diff
@@ -98,9 +98,19 @@ arm_unmount()
...
+	mount_msdosfs /dev/${DEV_EFI} ${STAGEDIR}/mnt
+	cp -r ${STAGEDIR}/mnt/efi ${STAGEDIR}/boot/msdos/efi
+	umount ${STAGEDIR}/mnt
+	mdconfig -d -u ${DEV_EFI}
+}
+
 echo -n ">>> Building arm image... "
 
 arm_install_uboot
+arm_install_efi
```
fichtner, line 113: There is an UEFI flag for this... https://github.com/opnsense/tools/blob/17fa565a1589d029169504b3414d9aea4532b193/build/vm.sh#L87-L88

https://github.com/opnsense/tools/blob/17fa565a1589d029169504b3414d9aea4532b193/Makefile#L84

I don't mind adding UEFI to arm by default. But it would be nice to be able to disable it.

build/common.sh
```diff
@@ -36,7 +36,12 @@ while getopts ${OPTS} OPT; do
 		export PORTSREFURL=${OPTARG}
 		;;
 	a)
-		export PRODUCT_TARGET=${OPTARG%%:*}
+		TARGET=${OPTARG%%:*}
+		if [ ${TARGET} == aarch64 ]; then
```
fichtner, line 40: what is being passed from the command line? Ideally we only call `make arm DEVICE=RPI3` and that would take care of all of this without the need of immediate input rewrite

device/RPI2.conf
fichtner: Whoops, I meant SRCREVISION

I'll commit a fix to access that from the device config.

fichtner: 316eb45bd7a

build/common.sh
yrzr: Hi there. There are two reasons that I put these codes in common scripts instead of device config files.
First of all, this is a common configuration for all ARM devices. So I don't want to redundant the codes to every config file.
Moreover, apart from some specific ARM devices, there are also generic ARM machines, such as QEMU and ESXi virtual machines. These machines are not listed in `tools/device`, and I don't think they should.

build/common.sh
yrzr: Thanks, I have reverted the commit.

build/arm.sh
yrzr: Okay, I have made EFI installing for arm default and optional.

build/common.sh
yrzr: Well, the `ARCH` got by `Makefile` is through command `uname -p`, and passed to `common.sh` as `PRODUCT_TARGET `.
On aarch64 machines, the result of `uname -p` is `aarch64`, while we need `arm64` as `PRODUCT_TARGET ` to build aarch64 targets.

fichtner:
hi @yrzr there are still some errors in the config files, I marked the relevant comments unresolved. I understand the need to make ARM64 generic, but if we can route our efforts through a device config it's easier to do it step by step than to change all at once and maybe some if/else trickery we don't need.

yrzr:
Hi @fichtner. I understand your concern. I have now created a new configuration file `device/ARM64-GENERIC.conf` to handle that.

device/RPI2.conf
fichtner: resolved comment wasn't resolved

device/RPI2.conf
```diff
@@ -1,13 +1,14 @@
...
+CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd${PRODUCT_ABI}/bin
```
fichtner, line 4:
```suggestion
CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd${SRCREVISION}/bin
```

device/BANANAPI.conf
```diff
@@ -3,15 +3,15 @@
 # http://www.banana-pi.org/r1.html
 
 export MAKE_ARGS_DEV="
-CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd11.2/bin
+CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd${PRODUCT_ABI}/bin
```
fichtner, line 6:
```suggestion
CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd${SRCREVISION}/bin
```

device/CLEARFOG.conf
```diff
@@ -0,0 +1,23 @@
+# https://www.solid-run.com/marvell-armada-family/clearfog/
+
+export MAKE_ARGS_DEV="
+CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd${PRODUCT_ABI}/bin
```
fichtner, line 4:
```suggestion
CROSS_BINUTILS_PREFIX=/usr/local/arm-gnueabi-freebsd${SRCREVISION}/bin
```

fichtner:
Merged, thanks!

PR #228, opened 2021-04-26
Add puppet-agent plugin

jan-win1993:
https://github.com/opnsense/plugins/pull/2358

### DIFF
--- a/config/21.1/plugins.conf
+++ b/config/21.1/plugins.conf
@@ -69,7 +69,7 @@ sysutils/lcdproc-sdeclcd			arm,arm64
 sysutils/munin-node				arm,arm64
 sysutils/node_exporter				arm,arm64
 sysutils/nut					arm,arm64
-sysutils/puppet-agent
+sysutils/puppet-agent				arm,arm64
 sysutils/smart					arm,arm64
 sysutils/virtualbox				arm,arm64
 sysutils/vmware					arm,arm64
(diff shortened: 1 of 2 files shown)


fraenki:
@fichtner This is not in 21.1.6 yet, right?

fichtner:
Yes, not enough time if we wanted to release 21.1.6 this week.

fichtner:
(maybe you have some time to talk on IRC, second day back at work for me)

config/21.1/ports.conf
```diff
@@ -221,6 +221,7 @@ sysutils/munin-node				arm,arm64
 sysutils/node_exporter				arm,arm64
 sysutils/nut					arm,arm64
 sysutils/pftop
+sysutils/puppet7
```
fichtner, line 224:
```suggestion
sysutils/puppet7					arm,arm64
```

config/21.1/plugins.conf
```diff
@@ -69,6 +69,7 @@ sysutils/lcdproc-sdeclcd			arm,arm64
 sysutils/munin-node				arm,arm64
 sysutils/node_exporter				arm,arm64
 sysutils/nut					arm,arm64
+sysutils/puppet-agent
```
fichtner, line 72:
```suggestion
sysutils/puppet-agent				arm,arm64
```

fichtner:
I'm getting a problematic error message here for the build during final puppet7 build:

    Could not load facter; cannot install

Have you guys ever seen this?

fichtner:
I think I found out why, geez 🤦

fichtner:
Adds about 10 MB to the packages and builds in 1-2 minutes so this is ok, thanks!

fichtner:
BTW, I'm bumping Ruby to 2.7 as per current FreeBSD ports default.

PR #234, opened 2021-06-08
Add py-opn-cli port

jan-win1993:


jan-win1993:
I've implemented the open recommendations from https://github.com/opnsense/plugins/pull/2358 and added the py-opn-cli dependency.

fichtner:
Merged, thanks!

PR #237, opened 2021-07-04
Add facility to create src.txz to build system

kraileth:
Due to the lack of the source distribution set, it is currently harder for people to build custom packages for their OPNsense installations than it has to be. Ideally the upcoming release could officially distribute src.txz.

This PR adds a simple script (basically base.sh modified accordingly). It also changes the Makefile to recognize the `src` target and add that as a dependency for `sets`. It's very simplistic (e.g. does no clean-up) as I didn't feel too confident with the build system.

I can try to make additional changes if requested, but it would probably easier if somebody actually familiar with the scripts would either do it or give me some hints on what to do.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@
 STEPS=		audit arm base boot chroot clean clone compress confirm core \
 		distfiles download dvd fingerprint info kernel list make.conf \
 		nano packages plugins ports prefetch print rebase release \
-		rename rewind serial sign skim test update upload verify \
+		rename rewind serial sign skim src test update upload verify \
 		vga vm xtools
 SCRIPTS=	batch distribution hotfix nightly
 
@@ -124,7 +124,7 @@ audit plugins: ports
 core: plugins
 packages test: core
 dvd nano serial vga vm: kernel core
-sets: kernel distfiles packages
+sets: kernel distfiles packages src
 images: dvd nano serial vga vm
 release: dvd nano serial vga
 
--- a/build/src.sh
+++ b/build/src.sh
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+# Copyright (c) 2014-2020 Franco Fichtner <franco@opnsense.org>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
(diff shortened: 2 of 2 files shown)


fichtner:
Hi @kraileth,

Pardon my asking, but what are typical consumers of src.txz? It looks like GitHub can easily extract a source tarball, i.e.:

https://github.com/opnsense/src/archive/refs/tags/21.1.8.tar.gz

OTOH, we have opnsense-code(8) tool that fetches source code from GitHub as well:

    # opnsense-code ports src

As for adding it to the build tools the ultimate question is if the set shall be signed and distributed by us, but for the aforementioned situation I don't know if that should be done.

I'm trying to be constructive and appreciate the contribution, but just feel there are details missing.


Cheers,
Franco

kraileth:
Hi @fichtner,

the solution that you pointed to here (using opnsense-code) is in fact what I've covered in [this](https://forum.opnsense.org/index.php?topic=21739.msg102499#msg102499) in February. It works fine but there are two reasons for my PR:

1) As OPNsense wants to closer track FreeBSD (switching away from HBSD, abandoning the Dfly installer, ...), I think having a src distset available makes sense. Installing the source with the OS has been an option with FreeBSD probably right from the start (I don't know for versions < 4.x).

2) OPNsense - in contrast to FreeBSD - is frequently used by avarage joe without a lot of *nix knowledge. Package building is not trivial by itself and I'm trying to cut down things and make them es easy as possible. I'd very much like to eliminate particularly the requirement of having to use Git to checkout the right tag. Maybe one day we can have a plugin that allows for graphical administration of package building; that's pie in the sky for sure, but everything that reduces the required steps and kills off some complexity certainly helps.

Well, that and the fact that it's not a terribly invasive change. Compared to many of the other make targets this one is neither time-consuming nor taking up a lot of space (source needs to be present for kernel and base, anyway).

And yes, preferably I'd like exactly what you mentioned: A signed source distset which makes the accompanying source for the respective version of OPNsense official.

AdSchellevis:
The tools repository more or less is our toolkit to build OPNsense, as we're not planning to use this, it's likely not being used by many and only causing noise in the process.  Fetching the source tarball from any of the  github repo's is still a viable option in our opinion. Currently we kindly have to decline.

PR #245, opened 2021-08-19
kernel: drop cross-build conditional for building kernel-toolchain

kevans91:
It's actually non-trivial to determine if you need to build the
kernel-toolchain or not -- host version not matching the target exactly
is a valid reason, among others.

Unconditionally build kernel-toolchain to err on the side of caution and
towards prevention of confusing build failures.  If the `base` target
has already been built then no building really happens here, we simply
walk through all of the bootstrap tools and determine they're all up to
date.

### DIFF
--- a/build/kernel.sh
+++ b/build/kernel.sh
@@ -58,9 +58,7 @@ __MAKE_CONF=
 ${MAKE_ARGS_DEV}
 "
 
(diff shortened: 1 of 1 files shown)


fichtner:
merged, thanks!

PR #247, opened 2021-09-12
config: switch to default loader flavor for 22.1

kevans91:
The core repo has already gained the appropriate brand and logo
definitions for lualoader, which appears to be the last obstacle to
dropping the forced-forthloader.

This has been tested with the stock lualoader from 21.7 and the equivalent to the following revisions applied; commits to stable/13 listed here for convenience:
- 0c675cb9c9cb ("lualoader: allow brands to specify a shift")
- 8df27f0acea8 ("lualoader: use more concise verbiage for autoboot")

(These were actually developed on opnsense, then forward-ported to FreeBSD-CURRENT and backported to stable branches :-))

Note that I've not tested that this all builds coherently because the build seems to kill my buildbox... but given that /boot/loader* is more-or-less identical to upstream, I'm guessing there's nothing else to this now that you've fixed the plist issue I caused over in opnsense/core.


fichtner:
Very cool, thanks a lot! I cherry-picked both changes to volatile/22.1 for the time being -- I have issues with a patch rebase after https://cgit.freebsd.org/src/commit/?h=stable/13&id=e8df60a69 that needs a bit more time to inspect.

PR #253, opened 2021-11-28
add HAProxy 2.4 

fraenki:
for https://github.com/opnsense/plugins/issues/2644

It would be good to have a package in order to be able to start testing.


fichtner:
Merged, thanks!

Maybe in the future we let the haproxy package in the configuration and just add the stable package as necessary?


Cheers,
Franco

fraenki:
@fichtner Good idea! :+1:

PR #256, opened 2021-12-24
device: cross-build is not always needed

yrzr:
cross-building is not needed when building aarch64 target on aarch64 host.


fichtner:
Looks reasonable. Will merge when I get hold of a computer in the next week.

fichtner:
@yrzr I changed your approach slightly for better integration in 3ed03c7223. couldn't really test at the moment so if you find an issue/typo I'm happy about a PR


Cheers,
Franco

fichtner:
On second thought it might be useful to have a flag PRODUCT_CROSS that is set when cross build is in progress? It might be useful later and avoids similar checks elsewhere in the code.

yrzr:
Everything looks fine. I have successfully built kernel with both native build and cross build.

Issue #258, opened 2021-12-27
22.1 ARM base build failed


yrzr:
Hi, I am trying to build 22.1 ARM with latest tools(3ed03c72237f0e12e06cb921c558824944529661) and src(1fca8e5780b58bdf99ede32bd0c9aaa23ef93e59). Both cross build on AMD64 or native build on ARM lead to the same error here:

```
--------------------------------------------------------------
>>> stage 4.4: building everything
--------------------------------------------------------------
...
/usr/local/aarch64-unknown-freebsd13.0/bin/ld: boot1.sym: error: PHDR segment not covered by LOAD segment
cc: error: linker command failed with exit code 1 (use -v to see invocation)
--- everything ---

make[1]: stopped in /usr/src
--- buildworld ---

make: stopped in /usr/src
*** Error code 2

Stop.
make: stopped in /usr/tools
```

Any ideas?

fichtner:
This might be an issue with llvm 13 update? Some things there have been reverted by FreeBSD already.

wilmardo:
Isn't this the error that comes from this bug? https://github.com/opnsense/tools/issues/250 

That is what I assumed, I am also having this error for at least the time that issue has been open.

fichtner:
No. This comes from FreeBSD source tree.

yrzr:
I have tried the FreeBSD source tree on branch stable/13, which results in the same problem.

fichtner:
It might be LLVM 13 which dropped in 22.1.b3. Wasn’t very eager about it but since we still follow stable/13 I don’t see a quick way out of this. Best to raise a bug at https://bugs.FreeBSD.org

wilmardo:
For reference, it seems that the bug has been opened:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=260972

TomRomeo:
Would it be possible to build 22.1.b2 on arm until this has been fixed upsteam?
I'm sorry if this his been asked before or if I simply missed the secion in the readme, but how exaclty do you downgrade to a stable release? Is it sufficient to just checkout the specific tag in /usr/tools?

fichtner:
Well, you just need to reset src.git:

    # git -C /usr/src reset —hard 22.1.b2

JPBeltman:
Got redirected to this issue by @fichtner on IRC. 

[...]

PR #266, opened 2022-02-09
device: fix `boot1.sym: error: PHDR segment not covered by LOAD segment`

yrzr:
fix problem in https://github.com/opnsense/tools/issues/258


fichtner:
I don't disagree but don't we need this for ports builds?

yrzr:
@fichtner Yes, I believe so. Let me rebuild and test first.

yrzr:
I have just finished a fresh native rebuild. As expected, it is safe to delete this on the native build.
On the other hand, cross-building takes much more time. Let's wait for the result.

fichtner:
@yrzr how about moving it to MAKE_ARGS_DEV_CROSS using 
e6e97ed ?

yrzr:
It would certainly work for native builds. But as I mentioned in https://github.com/opnsense/tools/issues/258, MAKE_ARGS_DEV_CROSS produces the same error in cross-builds. So I don't think this would solve the whole problem.

Moreover, I am cross-building without MAKE_ARGS_DEV_CROSS now. The kernel and base stages have been successfully made.  It is now working on the ports stage and has already produced about 289 packages. I guess it is also safe to remove MAKE_ARGS_DEV_CROSS on cross-builds. Let's wait for my final result.

fichtner:
Ok, then consider me convinced we can remove them all. Let’s keep MAKE_ARGS_DEV_CROSS support for later. It might be useful still.

yrzr:
Here is an update. I tried to cross-build as much as possible on AMD64 last week, without `CROSS_BINUTILS_PREFIX=/usr/local/aarch64-unknown-freebsd${SRCREVISION}/bin`.

As mentioned in https://github.com/opnsense/tools/issues/258#issuecomment-1039760120, `lang/go` fails in cross-builds. Therefore, I am forced to ignore some packages that rely on `lang/go`, such as `net-mgmt/telegraf`, `dnscrypt-proxy`, `net/wireguard` and etc. Apart from this problem, the building steps work well. I successfully get `base-22.1-aarch64.txz`, `kernel-22.1-aarch64.txz` and `packages-22.1-OpenSSL-aarch64.tar`.

Another problem is encountered when building the VM image. 

```bash
# make -C /usr/tools VERSION=22.1 DEVICE=ARM64 vm
......
>>> Begin extra: vm_hook
+ vm_hook /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64/mnt
+ loader_conf_fixup /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64/mnt
+ [ -f /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64/mnt/usr/local/etc/rc.loader ]
+ chroot /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64/mnt /usr/local/etc/rc.loader
chroot: /usr/local/etc/rc.loader: No such file or directory
*** Error code 1

Stop.
make: stopped in /usr/tools
```

I guess it is caused by chrooting into aarch64 environment without involving QEMU.

Therefore, I copied the mid-products to a native aarch64 build environment for VM image building. And everything works as expected.



So hereby I think we can finally come to this conclusion that`CROSS_BINUTILS_PREFIX=/usr/local/aarch64-unknown-freebsd${SRCREVISION}/bin` could be safely removed.

fichtner:
ok for me but your pr has conflicts

fichtner:
something went wrong so I merged this manually as 7680ca161a

PR #271, opened 2022-02-13
build/arm.sh: resize ARMSIZE to 2G

yrzr:
2G is enough as it takes about 1.8G in total.
Reduced size could help to save time in image-making, compressing, downloading, and disk writing.

### DIFF
--- a/build/arm.sh
+++ b/build/arm.sh
@@ -39,7 +39,7 @@ fi
 
 check_image ${SELF} ${@}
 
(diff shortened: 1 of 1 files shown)


fichtner:
This would deter the ability to do updates. You can pass the side to the script already if needed, but the minimum viability is 3.

yrzr:
Yes, I agree, The size could be passed to the scipts.

PR #279, opened 2022-03-09
add check_mk_agent

gratuxri:
check-mk-agent is needed by check-mk for monitoring freebsd instances.


ITJamie:
+1 on seeing this added

PR #289, opened 2022-05-09
Add IPMI support to Zabbix Proxy

mimugmail:
https://forum.opnsense.org/index.php?topic=28289.0

https://github.com/opnsense/ports/blob/master/net-mgmt/zabbix6-server/Makefile#L68


fichtner:
I have to take a quick look at openimpi... ;(

PR #291, opened 2022-05-17
Add build support for RockPro64

TheRandMan:
Adds a device config file for RockPro64.

### DIFF
--- a/device/ROCKPRO64.conf
+++ b/device/ROCKPRO64.conf
@@ -15,7 +15,7 @@ export PRODUCT_COMSPEED=1500000
 
 export ARM_FAT_SIZE="32m -b 16m"
 
-export ARM_UBOOT_DIR="/usr/local/share/u-boot/u-boot-rockpro64/"
+export ARM_UBOOT_DIR="/usr/local/share/u-boot/u-boot-rockpro64"
 
 arm_install_uboot()
 {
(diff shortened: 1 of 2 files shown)


device/ROCKPRO64.conf
```diff
@@ -0,0 +1,25 @@
...
+UBLDR_LOADADDR=0x42000000
+WITHOUT_MODULES=cloudabi32
+"
+
+export PRODUCT_KERNEL=SMP-ARM
+export PRODUCT_TARGET=arm64
+export PRODUCT_ARCH=aarch64
+export PRODUCT_WANTS="u-boot-rockpro64"
+export PRODUCT_WANTS_CROSS="aarch64-binutils qemu-user-static"
+export PRODUCT_COMSPEED=1500000
```
fichtner, line 14: That reminds me we should support this in the GUI too :)

device/ROCKPRO64.conf
```diff
@@ -0,0 +1,25 @@
...
+
+export ARM_FAT_SIZE="32m -b 16m"
+
+export ARM_UBOOT_DIR="/usr/local/share/u-boot/u-boot-rockpro64/"
```
fichtner, line 18: Note to self: trailing slash not needed

fichtner:
```suggestion
export ARM_UBOOT_DIR="/usr/local/share/u-boot/u-boot-rockpro64"
```

fichtner:
Merged, thanks!

device/ROCKPRO64.conf
fichtner: FYI https://github.com/opnsense/core/commit/7b1f3d70450

PR #292, opened 2022-05-19
Fix build failure during arm_install_uboot() on RockPro64

TheRandMan:
Build would fail when writing to the image's MSDOSBOOT partition without 'sysctl kern.geom.debugflags=0x10'


fichtner:
merged, thanks!

PR #294, opened 2022-06-01
distfiles target needs SRCDIR when ROOTDIR is not /usr/src

DSP1729:
Some ports need kernel sources, the source directory can be passed
in through SRC_BASE to the ports makefile. When ROORDIR is anything
but /usr/src this is needed for the distfiles target. Pass it in when the
package makefile runs.


fichtner:
@DSP1729 why not add to MAKE_ARGS ?

fichtner:
It didn't work, possibly broke https://github.com/opnsense/tools/issues/330

Issue #303, opened 2022-07-19
22.1 build: aarch64 failure of nut


turboproc:
Building master (native aarch64 build) results in failure when building the nut package. Relevant log below.
````
+ PKGLIST=nut
+ echo '>>> Installing packages in /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64: nut'
>>> Installing packages in /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64: nut
+ pkg -c /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64 remove -fya
Updating database digests format: 100%
Checking integrity... done (0 conflicting)
The following package(s) are locked and may not be removed:

        ca_root_nss
        curl
        expat
        git
        indexinfo
        libffi
        libidn2
        libnghttp2
        libpsl
        libunistring
        mpdecimal
        openssl
        p5-Error
        pcre2
        perl5
        pkg
        python39
        readline

Deinstallation has been requested for the following 1 packages (of 0 packages in the universe):

Installed packages to be REMOVED:
        node_exporter: 1.3.1_2

Number of packages to be removed: 1

The operation will free 10 MiB.
[1/1] Deinstalling node_exporter-1.3.1_2...
[1/1] Deleting files for node_exporter-1.3.1_2: 100%
+ echo pkg
+ sed 's/[^*]*//'
+ [ -n '' ]
+ PKGFOUND=''
+ cd /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64
+ find ./.pkg/All -name 'pkg-[0-9]*.pkg'
+ pkg -c /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64 add ./.pkg/All/pkg-1.17.5_1.pkg
Installing pkg-1.17.5_1...
the most recent version of pkg-1.17.5_1 is already installed
+ PKGFOUND=1
+ [ -z 1 ]
+ echo nut
+ sed 's/[^*]*//'
+ [ -n '' ]
+ PKGFOUND=''
+ cd /usr/obj/usr/tools/config/22.1/OpenSSL:aarch64
+ find ./.pkg/All -name 'nut-[0-9]*.pkg'
+ [ -z '' ]
+ echo 'Could not find package: nut'
Could not find package: nut
+ exit 1
*** Error code 1
```
Any idea where to fix this (previous releases had happy building on this platform). BASE and KERNEL have been built succesfully.

fichtner:
[...]

PR #304, opened 2022-07-20
Fix aarch64 build - disables build of node_exporter and nut packages

turboproc:
Fixes build on aarch64 by disable node_exporter and nut. Building
these packages would fail due to freeipmi unavailability.

Fixes: #303


fichtner:
22.1 directory no longer exists. 22.7 comes out next week :)

PR #305, opened 2022-07-20
Fix aarch64 build

turboproc:
Fixes build on aarch64 by disable node_exporter and nut. Building
these packages would fail due to freeipmi unavailability.

Fixes: #303


fichtner:
Merged, thanks!

PR #311, opened 2022-08-01
add zabbix6.2 to ports.conf

clanto007:


fichtner:
already done (and this patch was incomplete)

PR #312, opened 2022-08-04
FreeBSD 13.1 instructions

blkmajik:
Add documentation on how to resolve the pkg problem when starting with a new fresh FreeBSD 13.1 install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ and at least 4GB of RAM to successfully build all standard images.
 All tasks require a root user.
 
 
-PKG Shenanigans
----------------
+*pkg* Shenanigans
+-----------------
 
 Upstream keeps making incompatible changes to ``pkg`` which causes build
 failures.  In order to work around this problem you must use the OPNsense
@@ -29,18 +29,37 @@ jail pkg.
 To work around this you will need to use OPNsense as your package repository
 instead of FreeBSD.
 
-Create directory structure:
+First install git and acquire the `core` repository.  Please note that this
+portion is using the upstream FreeBSD packages:
 
-    # mkdir -p /usr/local/etc/pkg/repos/
-    # mkdir -p /usr/local/etc/pkg/fingerprints/OPNSense/trusted/
+    # pkg install git
+    # cd /usr
+    # git clone https://github.com/opnsense/core.git
+    # git checkout origin/stable/22.7
 
-In the `/usr/local/etc/pkg/repos/` directory you'll need to create two files.
-The first file is `FreeBSD.conf`:
+Make a backup copy of the FreeBSD pkg-static to use later:
 
-    FreeBSD: {enabled: no}
+    # cp /usr/local/sbin/pkg-static /root/pkg-static
 
-The second file is `OPNsense.conf`.  The `url` line is version specific
-so make sure it reflects your build platform:
+Remove all existing packages.  *The `-f` flag is required to delete `pkg`
+itself*:
+
+    # /root/pkg-static info | awk '{print $1}' | \
+        xargs /root/pkg-static delete -fy
+
+Create necessary directory structure using the `core` git repo we checked
+out earlier:
+
+    # cp -a /usr/core/src/etc/pkg /usr/local/etc/
+
+Change into the `/usr/local/etc/pkg/repos/` directory for these next steps.
+First you will need to disable the FreeBSD package repository:
+
+    # cd /usr/local/etc/pkg/repos
+    # mv FreeBSD.conf.shadow FreeBSD.conf
+
+Second you will need to create the `OPNsense.conf` file.  Use the following
+contents for 22.7 on amd64.
 
     OPNsense: {
         fingerprints: "/usr/local/etc/pkg/fingerprints/OPNsense",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,11 +7,81 @@ create sets, packages and images for the OPNsense project.
 Setting up a build system
 =========================
 
-Install [FreeBSD](https://www.freebsd.org/) 13.0-RELEASE for amd64
+Install [FreeBSD](https://www.freebsd.org/) 13.1-RELEASE for amd64
 on a machine with at least 25GB of hard disk (UFS works better than ZFS)
 and at least 4GB of RAM to successfully build all standard images.
-All tasks require a root user.  Do the following to grab the repositories
-(overwriting standard ports and src):
+All tasks require a root user.
```
fichtner, line 13: Maybe this can go below as well, seems a bit lost here.

README.md
```diff
@@ -7,11 +7,81 @@ create sets, packages and images for the OPNsense project.
...
+
+This is necessary because OPNsense builds within a jail and most but not
+all operations happen with the jails version of `pkg`. There are some aspects
+of the build process that operate outside the jail and those steps require
+interoperability between the base pkg and the jail pkg.
+
+To enable this compatibility you will need to use the OPNsense package
+repositories instead of the FreeBSD ones.
+
+First install git and acquire the `core` repository.  Please note that this
```
fichtner, line 32: I'm a little doubting if we have to make all of these instructions vs. simply adding a "fix" make target that runs the required sequence and can be easily tested and leaves almost no room for error? :)

README.md
blkmajik: This is now down with the original instructions.

README.md
blkmajik: I've created a 'make fix' target that should be version independent and with with the variables from the Makefile.

blkmajik:
OK I've updated this based on the previous feedback.  I now add a "fix" target to the Makefile that should do everything from my instructions in a version independent way.  With this the instructions have become simpler.   

I'm not 100% attached to using /root/pkg-static as a target but it is a hack that works around this situation.  There is likely a better place to put the pkg-static version that we need temporarily.

wilmardo:
1 just tested this but it might be nice to add a warning that this will remove all previous installed packages. This is the list that it removed:

```
Installed packages to be REMOVED:
        aarch64-binutils: 2.37_2,1
        ca_root_nss: 3.78
        curl: 7.84.0
        dialog4ports: 0.1.6_1
        expat: 2.4.8
        gettext-runtime: 0.21
        git: 2.37.1
        glib: 2.72.2,2
        indexinfo: 0.3.1
        libevent: 2.1.12
        libffi: 3.4.2
        libiconv: 1.16
        libidn2: 2.3.2
        libnghttp2: 1.48.0
        libpsl: 0.21.1_4
        libssh2: 1.10.0,3
        libunistring: 1.0
        libxml2: 2.9.13_2
        mpdecimal: 2.5.1
        p5-Authen-SASL: 2.16_1
        p5-CGI: 4.54
        p5-Clone: 0.45
        p5-Digest-HMAC: 1.04
        p5-Encode-Locale: 1.05
        p5-Error: 0.17029
        p5-GSSAPI: 0.28_2
        p5-HTML-Parser: 3.78
        p5-HTML-Tagset: 3.20_1
        p5-HTTP-Date: 6.05
        p5-HTTP-Message: 6.37
        p5-IO-HTML: 1.004
        p5-IO-Socket-INET6: 2.72_1
        p5-IO-Socket-SSL: 2.074
        p5-LWP-MediaTypes: 6.04
        p5-Mozilla-CA: 20211001
        p5-Net-SSLeay: 1.92
        p5-Socket6: 0.29
        p5-TimeDate: 2.33,1
        p5-URI: 5.10
        pcre: 8.45_1
        pcre2: 10.40
        perl5: 5.32.1_1
        python38: 3.8.13_2
        python39: 3.9.13
        qemu-guest-agent: 5.0.1
        qemu-user-static: 3.1.0_13
        readline: 8.1.2
        tmux: 3.2a
```

Not really a big deal for me since it was a build machine but it might be nice to add a warning and a `are you sure` prompt.

rmundel:
Just tested the fix and it seems to work.

Makefile
```diff
@@ -193,3 +193,34 @@ _OS:=	${_OS:C/-.*//}
...
+	pkg install -y git
+
+/root/pkg-static:
+	cp /usr/local/sbin/pkg-static /root/pkg-static
+
+fix: /usr/local/bin/git update /root/pkg-static
+	/root/pkg-static info | \
+		awk '{print $$1}' | \
+		xargs /root/pkg-static delete -fy
+	cp -a ${COREDIR}/src/etc/pkg /usr/local/etc/
```
jkuettner, line 212: According to the script, the `pkg` directory should still exist at the destination, right? Then the last slash is too much:

```suggestion
	cp -a ${COREDIR}/src/etc/pkg /usr/local/etc
```

jkuettner:
Additionally, I noticed that it is no longer possible to install dependencies for cross-compiling afterwards (e.g. `u-boot-rpi4` or `rpi-firmware`).

To achieve this, all required packages that are not available in the opnsense repo must be downloaded beforehand. For the example of `DEVICE=RPI` this would be the following steps:

```shell
# pkg fetch u-boot-rpi4 rpi-firmware aarch64-binutils qemu-user-static
...
# make fix
...
# make update
...
# cd /var/cache/pkg/
# pkg add u-boot-rpi4-2022.04_1~5127dabef1.pkg
# pkg add rpi-firmware-1.20210303.g20210303~4612691249.pkg
# pkg add aarch64-binutils-2.37_4,1~c3d50b5d6e.pkg
# pkg add qemu-user-static-3.1.0_13~a1d171ce8d.pkg
...
# cd /usr/tools
# make base kernel DEVICE=BANANAPI
...
```

fichtner:
I don’t think it helps to over complicate this PR with unrelated things. PROUCT_WANTS still exist. If there is an issue with it it should be addressed separately.

As for this PR we will be switching to pkg 1.18 shortly so it becomes obsolete. Discussed merging this internally, but we don’t want temporary workarounds cluttering the repository that are outside of our control.

fichtner:
pkg 1.18.4 is now in testing... in hindsight bd63e02c87f was and will still be very effective in the future.

PR #316, opened 2022-08-22
Allow building Arm images without cross compilation

JohannesNeu:
qemu-user-static is as of now not available on Freebsd13.1 aarch64 and it is not needed, because you do not need to cross compile in this case. Without this change you would need to modify RPI.conf by hand.

### DIFF
--- a/device/ARM64.conf
+++ b/device/ARM64.conf
@@ -8,7 +8,12 @@ WITHOUT_MODULES=cloudabi32
 export PRODUCT_KERNEL=SMP-ARM
 export PRODUCT_TARGET=arm64
 export PRODUCT_ARCH=aarch64
-export PRODUCT_WANTS_CROSS="aarch64-binutils qemu-user-static"
+export PRODUCT_WANTS_CROSS=$(if [ ${PRODUCT_HOST} != ${PRODUCT_ARCH} ]; then
+                                 echo -n "aarch64-binutils qemu-user-static"
+                             else
+                                 echo -n "aarch64-binutils"
+                             fi
+                            )
 
 # unset this for generic device handling, i.e. no device suffix
 unset PRODUCT_DEVICE
--- a/device/RPI.conf
+++ b/device/RPI.conf
@@ -11,7 +11,12 @@ export PRODUCT_KERNEL=SMP-ARM
 export PRODUCT_TARGET=arm64
 export PRODUCT_ARCH=aarch64
 export PRODUCT_WANTS="u-boot-rpi4 rpi-firmware"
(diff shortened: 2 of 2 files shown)


fichtner:
TBH, it really looks like the wrong fix: if _CROSS is already selected it's a cross-build so it indicates _CROSS is selected even when not cross-building.

fichtner:
And if this is about qemu-user-static removal there might be a more generic way to do it...

```
% git grep qemu-user-static
device/ARM64.conf:export PRODUCT_WANTS_CROSS="aarch64-binutils qemu-user-static"
device/BANANAPI.conf:export PRODUCT_WANTS_CROSS="arm-none-eabi-binutils qemu-user-static"
device/CLEARFOG.conf:export PRODUCT_WANTS_CROSS="arm-none-eabi-binutils qemu-user-static"
device/R4S.conf:export PRODUCT_WANTS_CROSS="aarch64-binutils qemu-user-static"
device/ROCKPRO64.conf:export PRODUCT_WANTS_CROSS="aarch64-binutils qemu-user-static"
device/RPI.conf:export PRODUCT_WANTS_CROSS="aarch64-binutils qemu-user-static"
device/RPI2.conf:export PRODUCT_WANTS_CROSS="arm-none-eabi-binutils qemu-user-static"
```

JohannesNeu:
Thanks for the reply, maybe you can elaborate a bit about the tools setup?
I am not that familiar with it.
Where is the _CROSS usually selected? I have never explicitly set a flag, when I compiled it in the past. I just set the DEVICE=RPI parameter. So I assume there is another script that does this?
Where would you suggest to put this fix?

fichtner:
See

https://github.com/opnsense/tools/blob/75b01e9874c17fd90a7fabbd5a4cc368472c833e/build/common.sh#L295-L299

JohannesNeu:
mhm, this is odd, the check in common.sh and the one I wrote down looks pretty much the same:
`if [ ${PRODUCT_HOST} != ${PRODUCT_ARCH} ]; then `
and
`(if [ ${PRODUCT_HOST} != ${PRODUCT_ARCH} ]; then`
I probably even copied it from there and modified it a bit.
I do not understand why this check would not work in common.sh and work if written to the RPI.conf.
Does the RPI.conf overwrite the exports in common.sh if I set DEVICE=RPI ?
Are the exports in RPI.conf done after the check in common.sh?

yrzr:
@JohannesNeu I have been building from native aarch64 for a long time using commands like `make -C /usr/tools 'VERSION=23.7.r3' 'DEVICE=ARM64' base`. It is being handled very well and seems not be a problem.

fichtner:
Going to close this as a timeout.

PR #322, opened 2022-10-18
build: Allow delimiters other than newline in {AUX,PLUGINS,PORTS}LIST

johnnyslee:
Currently, when assigning multiple items to a LIST, the following works:
```
make ports PORTSLIST="port/a
port/b"
```
while the following would fail:
```
make ports PORTSLIST="port/a port/b"
```

This patch makes comma, space and tab acceptable as delimiters instead of newline only.

### DIFF
--- a/build/audit.sh
+++ b/build/audit.sh
@@ -31,9 +31,7 @@ SELF=audit
 
 . ./common.sh
 
-if [ -z "${PORTSLIST}" ]; then
-	PORTSLIST=$(list_packages ${CONFIGDIR}/ports.conf)
-fi
+PORTSLIST=$(list_packages "${PORTSLIST}" ${CONFIGDIR}/ports.conf)
 
 setup_stage ${STAGEDIR}
 setup_base ${STAGEDIR}
--- a/build/common.sh
+++ b/build/common.sh
@@ -1233,5 +1233,10 @@ list_packages()
 {
 	local LIST_MATCH=1
 
-	list_config ${@}
+	if [ -n "${1}" ]; then
+		echo "${1}" | tr ',[:blank:]' '\n'
+	else
+		shift
+		list_config ${@}
+	fi
 }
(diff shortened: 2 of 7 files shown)


build/audit.sh
```diff
@@ -31,7 +31,9 @@ SELF=audit
 
 . ./common.sh
 
-if [ -z "${PORTSLIST}" ]; then
+if [ "${PORTSLIST}" ]; then
```
fichtner, line 34: that should be:

```suggestion
if [ -n "${PORTSLIST}" ]; then
```

I'm somewhat annoyed at the code growth just to handle external parameter passing and keeping it flexible. Possibly a better solution would be to bounce the configuration files through STAGEDIR?

build/common.sh
```diff
@@ -1235,3 +1235,8 @@ list_packages()
 
 	list_config ${@}
 }
+
+multi_line_list()
+{
+	echo "${@}" | tr ',[:blank:]' '\n'
```
fichtner, line 1241: the old function had a for loop which is softer to the eye and does auto-conversion of whitespace separated entities. but in any case this is so generic that it hardly makes sense to unify.. old code folded all in list_config/list_packages (or it's former equivalents)

build/audit.sh
```diff
@@ -31,7 +31,9 @@ SELF=audit
...
+	PORTSLIST=$(multi_line_list "${PORTSLIST}")
+else
 	PORTSLIST=$(list_packages ${CONFIGDIR}/ports.conf)
```
fichtner, line 37: Sorry to derail this... how about making this simpler? we would save the full if construct and could push conversion down to list_packages() where it used to be?

```suggestion
	PORTSLIST=$(list_packages "${PORTSLIST}" ${CONFIGDIR}/ports.conf)
```

fichtner:
Merged, thanks!

fichtner:
FYI cb807b82aeb735 hopefully this time I didn't break it...  ;)

PR #324, opened 2022-10-21
net/haproxy: permanently enable Prometheus exporter

fraenki:
In https://github.com/opnsense/plugins/issues/2764 official support for the Prometheus exporter was added to the HAProxy plugin. Now the HAProxy port must always be build with PROMEX enabled, otherwise the HAProxy plugin would stop working for all user of the Prometheus exporter.

Although PROMEX is currently enabled by default in the upstream port, we should ensure for this and all future versions of HAProxy that it remains enabled, no matter how the upstream port evolves.


fichtner:
Hmm, then this will generate an error in the `options` build step? Usually I try to handle this though port maintenance is unpredictable in renaming and modifying options…

fraenki:
I've proposed to add this, because at least I am incapable to remember all the required options for all ports. So instead of relying on the default options to always provide what we need, I tend to always set it explicitely. But yes, this will result in an error, if the option is renamed in upstream ports.

But I'm fine with whatever works best for you. :)

config/22.7/make.conf
```diff
@@ -78,7 +78,7 @@ net_asterisk16_UNSET=		DAHDI XMPP
 net_freeradius3_SET=		LDAP MITKRB_PORT MYSQL SQLITE3
 net_freeradius3_UNSET=		HEIMDAL
 net_frr7_SET=			MULTIPATH SNMP
-net_haproxy24_SET=		LUA
+net_haproxy24_SET=		LUA PROMEX
```
fichtner, line 81:
```suggestion
net_haproxy24_SET=		LUA # PROMEX
```

maybe if we settle on this you can review this as necessary and the script is silent?

fraenki: LGTM :)

fichtner:
Merged, thanks!

Issue #328, opened 2022-11-18
Add any mqtt client (like mosquitto)


Fuzail98:
IoT is like the most dynamic tech in the world today. And MQTT protocol is being used a lot. OPNsense is also being used like rapid fire. So requesting to re-add the mosquitto package or add any other mqtt client package and/or plugins.

Thanks in advance!

PR #329, opened 2022-11-21
Adjust make args for mosquitto port

mimugmail:
Related to #328 .. I'd remove the one in my repo then, it seems this option is kind of use ...

https://github.com/mimugmail/opn-repo/issues/16


fichtner:
merged, thanks!

PR #331, opened 2022-12-30
security/tailscale

mimugmail:
Add tailscale to the show .. time to add a plugin for it :)

### DIFF
--- a/config/23.1/ports.conf
+++ b/config/23.1/ports.conf
@@ -201,6 +201,7 @@ security/strongswan
 security/stunnel				arm
 security/sudo
 security/suricata				arm,aarch64
+security/tailscale				arm,aarch64
 security/tinc					arm
 security/tor					arm
 security/vault					arm
--- a/config/22.7/ports.conf
+++ b/config/22.7/ports.conf
@@ -201,6 +201,7 @@ security/strongswan
 security/stunnel				arm
 security/sudo
 security/suricata				arm,aarch64
+security/tailscale				arm,aarch64
 security/tinc					arm
 security/tor					arm,LibreSSL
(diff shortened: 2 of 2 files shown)


cosmedd:
+1 on seeing this added

config/23.1/ports.conf
```diff
@@ -201,6 +201,7 @@ security/strongswan
 security/stunnel				arm
 security/sudo
 security/suricata				arm,aarch64
+security/tailscale				arm,aarch64
```
yrzr, line 204: Why ignore arm and aarch64?

https://www.freshports.org/security/tailscale/ shows it works on both arches.

fichtner: usually just to not pollute build time for others... also because build has not been verified as we 99% deal with amd64

fichtner:
this is still open as it hasn't been discussed internally and we like to avoid GUIs on top of GUIs when we can

yrzr:
> this is still open as it hasn't been discussed internally and we like to avoid GUIs on top of GUIs when we can

Are you planning to make a new plugin for it?

fichtner:
well that would be the main question to discuss ;)

yrzr:
I vote for yes ;)

fichtner:
If we neither have funding nor a customer with a use case I think this is going to flop, because we can't just add stuff. I expect this to be a couple of days work at least.

fichtner:
So tailscale is aware and we proposed a cooperation a while back but at the moment there is no response from them so this stays parked right here.

fichtner:
0c2f10b26b6ed9ff7c9b8202b6413bdfc6985cc3

PR #332, opened 2023-01-04
add HAProxy 2.6 on OPNsense 23.1

fraenki:
required for https://github.com/opnsense/plugins/issues/3026


fichtner:
my bad for not seeing this!

with the merge to master we need both plugins on 22.7 unfortunately

fichtner:
let me handle the other thing. thanks!

fraenki:
thanks :)

fichtner:
@fraenki FYI, net/haproxy26 was branched and net/haproxy was updated to 2.7 -- I'm holding back on updating net/haproxy in your interest, but we should be switching to net/haproxy26 in 23.1.1 if you can find the time.


Cheers,
Franco

PR #357, opened 2023-06-01
config: Add dns/inadyn ddns agent

rhyser9:


fichtner:
At the moment we have been extending the native backend and ddclient is also back on track with releases so I have to decline this. Sorry.

PR #358, opened 2023-06-07
Add Zabbix 6.4

mimugmail:
Add Zabbix 6.4


fichtner:
It's likely missing make.conf bits

PR #361, opened 2023-07-14
enable packages that can be built on aarch64

yrzr:
see https://ftp.yrzr.tk/opnsense/FreeBSD:13:aarch64/23.1/MINT/23.1.11/All/

### DIFF
--- a/config/23.7/ports.conf
+++ b/config/23.7/ports.conf
@@ -119,6 +119,7 @@ net/hostapd
 net/igmpproxy
 net/isc-dhcp44-relay
 net/isc-dhcp44-server
+net/kea
 net/librdkafka					arm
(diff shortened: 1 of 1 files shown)


config/23.1/plugins.conf
```diff
@@ -4,7 +4,7 @@ databases/redis					arm
 devel/debug					arm
 devel/grid_example
 devel/helloworld
-dns/bind					arm,aarch64
+dns/bind					arma
```
AdSchellevis, line 7: typo?

yrzr: Yeah...fixed

fichtner:
It would make more sense to do this on 23.7 config because 23.1 is going away really soon.

PR #369, opened 2023-07-26
Fix device-specific hooks not getting loaded

maurice-w:
https://github.com/opnsense/tools/pull/367#issuecomment-1652237576


fichtner:
Merged, thanks!

PR #371, opened 2023-08-04
Add devel/py-jmespath

mimugmail:
It's needed when doing some more in-depth queries with ansible.


mimugmail:
It's already in repo, maybe due to some other dependency ..

PR #379, opened 2023-09-20
Add devel/abseil and devel/re2

mmetc:
The libraries are now a required dependency for crowdsec v1.5.4, which has been released today and submitted to the upstream freebsd ports.

### DIFF
--- a/config/23.7/ports.conf
+++ b/config/23.7/ports.conf
@@ -19,6 +19,7 @@ databases/py-sqlite3@py${PRODUCT_PYTHON}
 databases/pymongo				arm
 databases/redis					arm
 databases/rrdtool
+devel/abseil
 devel/arcanist@php${PRODUCT_PHP}		arm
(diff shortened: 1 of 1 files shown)


fichtner:
They don't have to be added if properly hooked up in the existing crowdsec package?

mmetc:
I don't know if declaring them in LIB_DEPENDS is enough, because they are needed at runtime. If my patch to devel/abseil is accepted, I'll switch again to a static build.
If you say it works, thanks

mmetc:
I mean I don't know what exactly gets shipped in the opnsense repo, I though it had to be defined here

fichtner:
Let me see after the next ports sync and close or merge this, but I have a feeling it's not required :)

fichtner:
Doesn't look like it's required from tools.git side. Update will be in 23.7.5 as it is shipped in FreeBSD.

PR #382, opened 2023-10-10
net-mgmt/FusionInventory-Agent

mimugmail:
Add FusionInventory-Agent for tools like GLPI ( https://glpi-project.org/ )

Refernce: https://forum.opnsense.org/index.php?topic=9903.0


fichtner:
Thanks!

PR #386, opened 2023-11-18
Publish the checksums (also) for the unpacked image files

doktornotor:
Rationale: You should be able to check the sanity of the unpacked image as well, in case you use some weird broken unpacking tool. Reference: https://forum.opnsense.org/index.php?topic=37074.0

**Note: This patch is completely untested, really do not have the time for testing ATM.**

### DIFF
--- a/build/compress.sh
+++ b/build/compress.sh
@@ -36,6 +36,9 @@ for ARG in ${@}; do
 	arm|dvd|nano|serial|vga|vm)
 		for IMAGE in $(find ${IMAGESDIR} -name \
 		    "*-${ARG}-*" \! -name "*.bz2" \! -name "*.sig"); do
+			echo -n ">>> Checksumming uncompressed image ${IMAGE} for ${PRODUCT_RELEASE}... "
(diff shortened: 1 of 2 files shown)


fichtner:
The checksums are for download integrity. It would probably be more useful to have signatures for the files? But to be honest „decompress broken“ is really uncommon and people don’t expect to check for it? 🤨

doktornotor:
> It would probably be more useful to have signatures for the files?

Signatures for the uncompressed files? Well, TBH I never used these signatures. I simply do not remember the syntax.

> But to be honest „decompress broken“ is really uncommon and people don’t expect to check for it?

Well, normally the tool would check for CRC errors. Then, I recall this unfortunate bzip2 bug: https://bugs.launchpad.net/ubuntu/+source/bzip2/+bug/1834494
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=931278

There things happen even with widely used tools. 🤷‍♂️

fichtner:
Fair enough. I've moved the generated signatures in the release set away from the bz2 to the actual uncompressed files since they were already generated. I don't want to overload the (easily modifiable) checksum file. Not sure if OpenSSL 3 made verification using Base64 encoding possible now in a single step, but worth investigating.

PR #388, opened 2023-11-24
config: ignore freeipmi on aarch64

maurice-w:
Was added in 1aae3368facf313f8bd601a1fc28092aa7fc4444, but is broken on aarch64:
https://svnweb.freebsd.org/ports?view=revision&revision=536924
https://www.freshports.org/sysutils/freeipmi


fichtner:
Merged, thanks!

PR #397, opened 2024-02-22
Update make.conf for caddy-custom build to mitigate build errors

Monviech:
As explained here:
https://github.com/Monviech/os-caddy-plugin/issues/102

Change the build priority to `desec` and `porkbun` instead of `vultr`, since they have been especially requested in past issues of os-caddy-plugin.

The issue stems from `vultr` having different `libdns` dependencies than the other modules, making the build fail if it's included with certain other combinations of modules (e.g. `desec`).

### DIFF
--- a/config/24.1/make.conf
+++ b/config/24.1/make.conf
@@ -105,12 +105,15 @@ CADDY_CUSTOM_PLUGINS=		github.com/caddyserver/ntlm-transport \
 				github.com/caddy-dns/googleclouddns \
 				github.com/caddy-dns/gandi \
 				github.com/caddy-dns/azure \
-				github.com/caddy-dns/vultr \
+				github.com/caddy-dns/porkbun \
 				github.com/caddy-dns/openstack-designate \
 				github.com/caddy-dns/google-domains \
 				github.com/caddy-dns/ovh \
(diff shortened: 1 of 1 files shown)


config/24.1/make.conf
```diff
@@ -123,8 +126,4 @@ CADDY_CUSTOM_PLUGINS=		github.com/caddyserver/ntlm-transport \
 				github.com/caddy-dns/hexonet \
 				github.com/caddy-dns/mailinabox
 
-CADDY_CUSTOM_PLUGINS_BROKEN=	github.com/caddy-dns/netcup \
-				github.com/caddy-dns/desec \
-				github.com/caddy-dns/porkbun \
-				github.com/caddy-dns/netlify \
-				github.com/caddy-dns/namesilo
+CADDY_CUSTOM_PLUGINS_BROKEN=	github.com/caddy-dns/vultr
```
fichtner, line 129: do you still need this? the _BROKEN thing I just put here for further reference but it has no effect and when you ship without vultr then we can drop it

Monviech: Removed mention of it, when it's fixed in the future I will evaluate the build including it again.

PR #400, opened 2024-03-24
www/caddy-custom: Make current build reproducible

Monviech:
Requires: https://github.com/opnsense/ports/pull/190

Lock the whole build down with commit hashes and version tag. Upstream changes won't make the build fail all of a sudden now (or at least mitigates most of these issues.)

### DIFF
--- a/config/24.1/make.conf
+++ b/config/24.1/make.conf
@@ -92,7 +92,6 @@ www_squid_UNSET=		AUTH_NIS TP_IPFW
 www_webgrind_SET=		CALLGRAPH
 
 # for www/caddy-custom
-CADDY_CUSTOM_VERSION=		v2.7.6
 CADDY_CUSTOM_PLUGINS=		github.com/caddyserver/ntlm-transport@e0c1e46a30093fa243d06a83964da5573ee6a51f \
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks!

Monviech:
@fichtner 

Remember this? https://github.com/opnsense/tools/pull/400/commits/214a924ce5867455b37532b44419a4e39b4e95b3

Now this happens:

2024/05/30 20:14:51 [INFO] Pinning versions
2024/05/30 20:14:51 [INFO] exec (timeout=0s): /usr/local/bin/go121 get -d -v -buildmode=exe -v -trimpath -ldflags=-s -buildvcs=false github.com/caddyserver/caddy/v2 
go: downloading github.com/caddyserver/caddy/v2 **v2.8.1**

Seems like it really downloads latest just like I assumed back then. I really want to pin the version to 2.7.6. That's also why the build failed here: https://forum.opnsense.org/index.php?topic=35828.msg200025#msg200025

The new Caddy version was juuust released: https://github.com/caddyserver/caddy/releases

fichtner:
Sure, but that's a FreeBSD ports fail. I think we know the fix if you make a PR in ports I can either try to report it to FreeBSD ports or you can try

PR #406, opened 2024-05-03
enable OFED userland

### LINKED ISSUE #405: OFED disabled in src.conf
Hi, I understand that yes, OPNsense does have the Infiniband drivers enabled by default, it doesn't seem to have OpenSM and other utilities from the OFED package required to run a Subnet Manager to use the full speed of IB cards. 

As it's not a port, I've had trouble building a package, and given the support for IB is already built into the kernel, is there a reason for the "WITHOUT_OFED=YES" in src.conf? 

I understand #65 was lodged in 2018 and closed later that year, but without OpenSM it's hard to get the full capabilities out of these cards, as currently with the standard 40Gbit interfaces the best that can be used on OPNsense is 10Gbit EoIB - as opposed to the full 40Gbit IPoIB/RDMA modes.

rileyg98:
Enable OFED userland binaries to be built

This fixes #405 as discussed


config/24.7/src.conf
```diff
@@ -20,7 +20,7 @@ WITHOUT_NLS=yes
 WITHOUT_NLS_CATALOGS=yes
 WITHOUT_NS_CACHING=yes
 WITHOUT_NTP=yes
-WITHOUT_OFED=yes
+WITH_OFED=yes
```
fichtner, line 23: I reckon this is the default so it can be omitted? ;)

rileyg98: It is the default on OpenBSD as I understand, but I'm happy to change that if you'd prefer.

rileyg98: I've pushed another change after this one to remove

fichtner: thanks, according to src.conf this is indeed the default on amd64.

fichtner:
Merged, thanks!

Issue #407, opened 2024-05-04
WITH_OFED_EXTRA also disabled


rileyg98:
Found that WITH_OFED only configures some of the userspace diagnostics - which are very helpful for configuration, but doesn't enable the actual subnet manager (without which an Infiniband network cannot run). As such, WITH_OFED_EXTRA is also required.

I will lodge a PR for this shortly.

fichtner:
I think you're looking for "opensm" specifically, right?

fichtner:
I get the feeling opensm is misplaced in WITH_OFED_EXTRA seeing it pulls a lot of diagnostics as well

rileyg98:
It seems so, but as I understand, ofed was upstreamed from Mellanox (now
nvidia) a while back and this is how they had it set up.

What's even more odd is that the opensm service is created with ofed extras
disabled, so it may be an error in upstream. I'm happy to go and lodge a
bug there, as I did find that someone mentioned there were changes back in
2021 and it may have caused a bug (where someone couldn't get opensm to
start).

I've been able to get it to work locally with a custom 24.1 build with the
extras enabled, and it doesn't seem to impact anything - except IPoIB
connections not being well understood by the GUI, but that's well out of
scope of the tools repo and something I may look at fixing myself down the
line in core.

Honestly this feels so much more like where a plugin or port would be
better, but for some reason it's in the base freeBSD build profile.

fichtner:
Thanks for the details.

I have a gut feeling that I'd rather want to change opensm to go with `WITH_OFED` as you mention that the extras don't build correctly... we could always let upstream fix but eventually it means handrolling their changes as well. It might not even be an upstream issue.

> I've been able to get it to work locally with a custom 24.1 build with the
> extras enabled, and it doesn't seem to impact anything - except IPoIB
> connections not being well understood by the GUI, but that's well out of
> scope of the tools repo and something I may look at fixing myself down the
> line in core.

That would be highly appreciated!


Cheers,
Franco

fichtner:
Previous PR binary result: https://github.com/opnsense/tools/commit/8190c8665a

opensm change: https://github.com/opnsense/src/commit/a1edc286c
Binary result: https://github.com/opnsense/tools/commit/de9c374e4

If that state works for you I'd avoid WITH_OFED_EXTRA for now.


Cheers,
Franco

rileyg98:
I'll test that tonight and have a look!

rileyg98:
[...]

PR #408, opened 2024-05-04
Add WITH_OFED_EXTRA for opensm

rileyg98:
WITH_OFED_EXTRA is required for opensm to be compiled

as per #407


fichtner:
Hehe, scope creep :) Let me check the current version first so I have a reference point on what comes in with this (also space-wise)

rileyg98:
Closing as unnecessary - de9c374 fixes with upstream.

PR #410, opened 2024-05-22
www/caddy-custom: Update dependencies of extra modules

Monviech:
* Fix some build errors that are introduced due to a conflict between different libdns versions. caddy-dns mod.ts introduce dependencies to their libdns counterpart, e.g. caddy-dns/porkbun depends on libdns/porkbun.

* This dependency has been replaced using a "go mod replace" command that is passed via the xcaddy command. It replaces the libdns module with a custom patched one. The line can be completely removed in the future, since when the build error of the dependency has been fixed, it can be pulled from upstream again.

* The regession that is fixed is that cloudflare has newer dependencies than the rest of the dns modules. That forces indirect upgrades of libdns versions in modules and some smaller build errors in these modules are introduced (namely, uint and int errors at a few spots).

* Two new dns modules have been added, netcup (per request) and rfc2136 as a generalized provider for any kind of dynamic DNS update.

### DIFF
--- a/config/24.1/make.conf
+++ b/config/24.1/make.conf
@@ -93,8 +93,8 @@ www_webgrind_SET=		CALLGRAPH
 
 # for www/caddy-custom
 CADDY_CUSTOM_PLUGINS=		github.com/caddyserver/ntlm-transport@e0c1e46a30093fa243d06a83964da5573ee6a51f \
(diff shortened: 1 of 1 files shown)


Monviech:
Upstream PRs:

https://github.com/libdns/desec/pull/5
https://github.com/libdns/porkbun/pull/6
https://github.com/libdns/netlify/pull/4
https://github.com/libdns/netcup/pull/1

fichtner:
Merged, thanks!

PR #413, opened 2024-05-30
www/caddy-custom: Remove rfc2136 since it causes build with caddy-2.8.1 to fail

Monviech:
The caddy-custom port just /pretends/ to build 2.7.6 (package is still called that way.) But it actually builds 2.8.1 now, very sneaky.

References:

https://github.com/opnsense/tools/pull/400#issuecomment-2140805870
https://forum.opnsense.org/index.php?topic=35828.msg200025#msg200025
https://forum.opnsense.org/index.php?topic=35828.msg200033#msg200033


fichtner:
Merged, thanks!

PR #414, opened 2024-06-03
www/caddy-custom: Update build for caddy-2.8.4, reintroduce rfc2136 with custom patch

Monviech:
- Update build for caddy-2.8.4, 
- Reintroduce rfc2136 with custom patch
- Update Porkbun and Netcup libdns dependency

**Attention, this build will fail with v2.7.6, it REQUIRES >=v2.8.0**


fichtner:
Merged, thanks!

PR #424, opened 2024-07-25
aarch64/ARM64 DVD creation

korhojoa:
Removes requirement for i386 cdboot for non-amd64 architectures. Allows aarch64 dvd target to build.

As many arm platforms support UEFI, having a bootable dvd image is convenient.

Before:

```sh
Populating `/usr/obj/usr/tools/config/24.1/aarch64/efiboot.img'
Image `/usr/obj/usr/tools/config/24.1/aarch64/efiboot.img' complete
>>> Building dvd image... makefs: cd9660_add_boot_disk: lstat("/usr/obj/usr/tools/config/24.1/aarch64/work/boot/cdboot"): No such file or directory
*** Error code 1

Stop.
make: stopped in /usr/tools/
```

After:
```sh
Populating `/usr/obj/usr/tools/config/24.1/aarch64/efiboot.img'
Image `/usr/obj/usr/tools/config/24.1/aarch64/efiboot.img' complete
>>> Building dvd image... done
```

### DIFF
--- a/build/dvd.sh
+++ b/build/dvd.sh
@@ -55,7 +55,8 @@ if [ -n "${PRODUCT_UEFI}" -a -z "${PRODUCT_UEFI%%*"${SELF}"*}" ]; then
 
 	setup_efiboot ${STAGEDIR}/efiboot.img \
 	    ${STAGEDIR}/work/boot/loader.efi 2048 12
-elif [ ${PRODUCT_ARCH} = "amd64" ]; then
+fi
+if [ ${PRODUCT_ARCH} = "amd64" ]; then
 	LEGACYBOOT="-o 'bootimage=i386;${STAGEDIR}/work/boot/cdboot' -o no-emul-boot"
 fi
 
--- a/build/dvd.sh
+++ b/build/dvd.sh
@@ -50,7 +50,11 @@ UEFIBOOT=
 LEGACYBOOT=
 
 if [ -n "${PRODUCT_UEFI}" -a -z "${PRODUCT_UEFI%%*"${SELF}"*}" ]; then
-	UEFIBOOT="-o bootimage=efi;${STAGEDIR}/efiboot.img"
+        if [ ${PRODUCT_ARCH} = "amd64" ]; then
+                UEFIBOOT="-o bootimage=i386;${STAGEDIR}/efiboot.img"
+        else
+                UEFIBOOT="-o bootimage=efi;${STAGEDIR}/efiboot.img"
+        fi
 	UEFIBOOT="${UEFIBOOT} -o no-emul-boot -o platformid=efi"
 
 	setup_efiboot ${STAGEDIR}/efiboot.img \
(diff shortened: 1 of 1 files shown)


fichtner:
I'm wondering if this is worth the inconsistency. DVD wasn't designed for ARM, which ARM devices sell with DVD drives? Or is this about VM? I'm assuming the resulting image was tested? Also I'm not sure we're not breaking amd64 here while at it.

korhojoa:
What's the inconsistency?

I have no idea about sales with DVD drives, but IPMI/IPKVM systems almost always provide a virtual drive to mount images with, and it's useful with that. (Then again, you can just add a usb dvd drive if you really want to burn a disc)

Yes, I tested an image created with this. It boots and installs fine on a hypervisor (UTM) and Raspberry Pi 5.
I'll try it with amd64 soon.

Currently the boot image is set for i386 to use cdboot (which doesn't exist for eg. aarch64). This just makes it not be set so EFI-based images can be built for other platforms.

fichtner:
first you're changing line 52 in amd64 UEFI mode, secondly legacy aarch64 seems to do nothing. hence the questions.

korhojoa:
Yes, legacyboot aarch64 doesn't do anything as it doesn't have a 80x86 (here i386) el torito boot mode.
I can rename it to i386 if the naming isn't helpful.

I've tested this on amd64 and on 24.7 now. Do you want example images somewhere or some kind of proof that they work?

Here's some screenshots if that helps, this is from EFI:
efi_boot

This is from i386/BIOS:
i386_boot

Here is 24.1 running on a Raspberry Pi 5, installed via the image generated:
Screenshot 2024-07-26 at 10-17-13 Dashboard Lobby OPNsense localdomain

Here is 24.7 running on UTM:
Screenshot 2024-07-26 at 10-20-39 Dashboard Lobby OPNsense localdomain

korhojoa:
Is there something more I can do for this?

I've built multiple images with amd64 and aarch64 with this change.
(On another note, having aarch64 packages and images officially available would be amazing. Many devices now run opnsense quite well.)

build/dvd.sh
```diff
@@ -47,14 +47,18 @@ setup_mtree ${STAGEDIR}/work
 setup_entropy ${STAGEDIR}/work
 
 UEFIBOOT=
+LEGACYBOOT=
 
 if [ -n "${PRODUCT_UEFI}" -a -z "${PRODUCT_UEFI%%*"${SELF}"*}" ]; then
-	UEFIBOOT="-o bootimage=i386;${STAGEDIR}/efiboot.img"
+	UEFIBOOT="-o bootimage=efi;${STAGEDIR}/efiboot.img"
```
fichtner, line 53: To comment on this again: I don't want to test this on amd64 for regression when I'm supposed to be building the next opnsense release and people wonder why it may not work as before.

korhojoa: Understandable, amd64 now has the sysid i386 also for the efi bootimage.

fichtner:
Here is my proposal based on your work. If you agree I will test it on Monday. https://github.com/opnsense/tools/pull/430

fichtner:
Closing as discussed.

PR #426, opened 2024-07-30
Add Rust to 24.7 aux set on aarch64

maurice-w:
https://forum.opnsense.org/index.php?topic=35828.msg205942#msg205942


fichtner:
Merged, thanks!

PR #430, opened 2024-08-30
build/dvd: only support UEFI build with amd64-based BIOS boot compat

fichtner:


fichtner:
Github added reactions because maintainers complained that too many +1 comments flooded their mails. So reactions do not send notifications.

In any case thanks. I will merge now and test next week.


Cheers,
Franco

maurice-w:
@korhojoa Could you please give this aarch64 DVD image a try? I made it using this patch.
https://opnsense-update.walker.earth/releases/24.7/

PR #432, opened 2024-09-16
www/caddy-custom: Add ratelimit module for CLI usage

Monviech:
Build in xcaddy checks out, doesn't hurt adding this for the CLI crowd who only use this port without GUI in OPNsense.

Fixes: https://github.com/opnsense/plugins/issues/4236


fichtner:
do we need a caddy-custom revision bump for this? thanks!

Monviech:
Do we control the revision now or is it still synced from upstream from the port?

fichtner:
Yes, both 😉

Monviech:
Sure lets do a revision bump then. Thank you.

fichtner:
👍

PR #435, opened 2024-09-27
fix: Add net-mgmt/xymon-client to ports.conf

### LINKED ISSUE #434: Add net-mgmt/xymon-client
Could we please add `net-mgmt/xymon-client`
https://freebsd.pkgs.org/14/freebsd-amd64/xymon-client-4.3.30.pkg.html
https://www.freshports.org/net-mgmt/xymon-client

Relates to opnsense/plugins#4259

> Xymon is a system for monitoring servers and networks. It has a great deal
of inspiration from the Big Brother monitor, but unlike Big Brother it is
designed to work well whether you need to monitor small network with just
a handful of hosts, or large networks with thousands of hosts.
Xymon is the successor to the bbgen toolkit, which has been available as
an add-on to Big Brother since late 2002. The name change was decided upon
when Xymon acquired enough functionality to be a stand-alone product.
The tools that formed the bbgen toolkit are still present in Xymon
and are quite important for it, so if you have used bbgen before,
Xymon will seem quite familiar.
This is the client.
LICENSE: GPL2 or later
Groups: xymon

kumy:
Closes #434

### DIFF
--- a/config/24.7/ports.conf
+++ b/config/24.7/ports.conf
@@ -96,6 +96,7 @@ net-mgmt/nrpe					arm
 net-mgmt/p5-FusionInventory-Agent
 net-mgmt/py-opn-cli
 net-mgmt/telegraf				arm
+net-mgmt/xymon-client				arm
 net-mgmt/yaf					arm
 net-mgmt/zabbix5-agent				arm
 net-mgmt/zabbix5-proxy				arm


config/24.7/ports.conf
```diff
@@ -96,6 +96,7 @@ net-mgmt/nrpe					arm
 net-mgmt/p5-FusionInventory-Agent
 net-mgmt/py-opn-cli
 net-mgmt/telegraf				arm
+net-mgmt/xymon-client
```
fichtner, line 99: for now I'd like to avoid "arm" build as well (that's the 32 bit arm, not aarch64)

fichtner:
@kumy thanks, I do need to check how much this adds to the build but in general the port looks ok

config/24.7/ports.conf
kumy: @fichtner done :heavy_check_mark:

kumy:
Done in https://github.com/opnsense/tools/commit/88e547d70eefde02137660f313ba079db4aeae68

PR #436, opened 2024-10-04
www/caddy-custom: Update layer4 module

Monviech:
Adds these two commits:

https://github.com/mholt/caddy-l4/commit/ea27408a3384a7f33b5d1cb49ec592cf76942be6
https://github.com/mholt/caddy-l4/commit/bdee6a6620c02aca4817e6c3091fed26aa10940c

Fixes (for now) that the TLS matcher in the Layer 4 module didn't work anymore due to some latest Chrome and Edge updates.

For tracking that issue:
https://github.com/mholt/caddy-l4/issues/250

PR: https://forum.opnsense.org/index.php?topic=42955.0


fichtner (approved):
By the time it is release it may be obsoleted already, but better to go forward. We also need a revision bump in ports caddy-custom.

Monviech:
Im unsure this will obsolete soon, since this issue was very obscure and only reproducible in FreeBSD. On linux this patch isnt needed, thats why its in its own branch. Im happy I got help to track it down and test it. Hope this will work for a while xD

PR #437, opened 2024-10-23
Update clean-obj target

cpalv:
currently, `make clean-obj` removes everything under `/usr/obj/`
only delete items under `/usr/obj${ROOTDIR}` since there might be other things there


fichtner:
Nope.

PR #439, opened 2024-10-25
Make it easier to copy/paste commands

tinyboxvk:


fichtner:
Hmm, I mean all it does is remove # which we designate to root shell usage because sometimes the context is unclear otherwise. I’d rather keep this to keep a unified style as this is used in a lot of places and also related to historic forum posts which can not really be changed.

tinyboxvk:
> Hmm, I mean all it does is remove #

I was thinking something like 

https://github.com/user-attachments/assets/8630e1f6-d08e-4669-8fe6-2ff6ea055c16

instead of

https://github.com/user-attachments/assets/f800ac85-6442-430b-8094-fd3b7cc61a45

One cannot easily copy and paste multi-line commands easily with # at the front of each line.

> because sometimes the context is unclear otherwise

I do agree with this.

PR #441, opened 2024-11-01
config/plugins: ignore Zenarmor vendor repository on aarch64

maurice-w:
Zenarmor is not available for OPNsense aarch64.
Bug report: https://forum.opnsense.org/index.php?topic=35828.msg218005#msg218005


fichtner:
Good idea, thanks!

PR #447, opened 2025-01-09
add haproxy 3.0

fraenki:
required for https://github.com/opnsense/plugins/pull/4458


fichtner:
Ok but we need both haproxy versions in both versions for the transition

fichtner:
I added net/haproxy to 24.7 additionally, maybe we can get away with avoiding 2.8 on 25.1 so I've kept it as submitted via 73ab164 -- thanks!

fraenki:
Thank you :)

PR #458, opened 2025-02-11
build/common: load device-specific environment before setting SRCABI

maurice-w:
`PRODUCT_ARCH` can be defined in device file, hence device file must be loaded first.

Currently, `make prefetch DEVICE=ARM64 MIRRORS=<custom mirror>` fails on amd64 build systems because `SRCABI` is incorrectly set to `FreeBSD:14:amd64`.

[Bug report by @mnaiman](https://github.com/maurice-w/opnsense-vm-images/issues/6#issuecomment-2648416749)

### DIFF
--- a/build/common.sh
+++ b/build/common.sh
@@ -233,7 +233,6 @@ SRCREVISION=unknown
 if [ -f ${SRCDIR}/sys/conf/newvers.sh ]; then
 	eval export SRC$(grep ^REVISION= ${SRCDIR}/sys/conf/newvers.sh)
 fi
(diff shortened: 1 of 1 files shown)


build/common.sh
```diff
@@ -242,6 +241,7 @@ fi
 
 # load device-specific environment
 . ${DEVICEDIR}/${PRODUCT_DEVICE_REAL}.conf
+export SRCABI="FreeBSD:${SRCREVISION%%.*}:${PRODUCT_ARCH}"
```
fichtner, line 244: I would move the whole block handing SRC* including the comment... that way it looks like it was meant to be ;)

maurice-w: Agreed, makes sense.

PR #459, opened 2025-02-13
build/vm: enable cross-building VM images

maurice-w:
Same procedure as in build/arm. Allows building aarch64 VM images on amd64 build systems.


fichtner:
Merged, thanks!

PR #460, opened 2025-02-13
device/*: cleanup MAKE_ARGS_DEV for arm / arm64 devices

maurice-w:
No need to exclude cloudabi32 from the build any longer since CloudABI was removed from FreeBSD back in 14.0.

https://man.freebsd.org/cgi/man.cgi?query=cloudabi&manpath=FreeBSD+13.4-RELEASE


fichtner:
Merged, thanks!

Issue #461, opened 2025-02-15
building readline-8.2.13_2 errors with code 71


computeronix:
When running the command 
make base kernel packages arm-3G DEVICE=RPI DEBUG=1

During the build process it eventually fails out during the process building readline-8.2.13_2 with the following error:

install: ./rltest2.c: No such file or directory
*** Error code 71

When manually building it errors out with this too.

Workaround that resolves the manually building:
```shell
cd /usr/ports/devel/readline
sudo pkg install devel/gmake devel/binutils devel/autoconf devel/libtool
sudo make config
sudo make
sudo make reinstall
```

I am not sure if this is expected to have these tools as part of building but if not maybe catch the error then mention this or check to enure these tools are installed and install them then attempt to build readline or re-build readline?

Environment is:
FreeBSD 14.2 (arm64)
Raspberry Pi 4

Thank you again as always!

computeronix:
I think I updated this PR properly, it looks like between aux.conf and ports.conf the only build package missing was devel/autoconf

so hopefully this resolves the readline issue for those without the full required packages above

computeronix:
devel/autoconf made no difference in build script after complete refresh of system - must of been lingering file

for arm64/aarch64 needed to manually install and run the workaround to get around the error code 71

follow the workaround / manual install path above if you get this

closing this out

```shell
cd /usr/ports/devel/readline
sudo pkg install devel/gmake devel/binutils devel/autoconf devel/libtool
sudo make config
sudo make
sudo make reinstall
```

PR #462, opened 2025-02-15
Add devel/autoconf support

computeronix:
This should resolve https://github.com/opnsense/tools/issues/461 where readliine does not fully build missing dependency issues?

### DIFF
--- a/config/25.1/aux.conf
+++ b/config/25.1/aux.conf
@@ -1,6 +1,7 @@
 #ORIGIN						IGNORE
 devel/binutils
 devel/cmake-core
+devel/autoconf
 lang/go121					arm
 lang/go122					arm
 lang/go123					arm


fichtner:
I don't mind removing the "arm" label for duckdb but the other things don't look like a fix for a problem we're having. If FreeBSD ports have trouble building on RPI it would be best to head over to https://bugs.freebsd.org

config/25.1/ports.conf
```diff
@@ -14,13 +14,14 @@ databases/p5-DBD-Sybase				arm
...
 databases/php${PRODUCT_PHP}-sqlite3
-databases/py-duckdb@py${PRODUCT_PYTHON}		arm
+databases/py-duckdb@py${PRODUCT_PYTHON}
 databases/py-redis@py${PRODUCT_PYTHON}		arm
 databases/py-sqlite3@py${PRODUCT_PYTHON}
 databases/pymongo				arm
 databases/redis72				arm
 databases/rrdtool
 devel/arcanist@php${PRODUCT_PHP}		arm
+devel/autoconf
```
fichtner, line 24: autoconf is a dependency of automake in the line below, this is sort of implied by the fact that https://pkg.opnsense.org/FreeBSD:14:amd64/25.1/MINT/25.1/latest/All/autoconf-2.72.pkg exists in the packages...

computeronix: interesting! let me see if I can triple check this and attempt to replicate it again as I didn't check to the dependency there and the workaround posted in the PR happened to work and the only package I noticed that was missing was devel/autoconf.  Will see if this acts differently for some reason and report as a bug to freebsd as suggested if so!

Thanks again for that too!

fichtner: in some cases the mere fact that you retry building after a failure can fix problems seen in the FreeBSD ports (and some of these build failures and reorders are probably arch dependant and not seen easily on amd64 which we do nightly builds on)

config/25.1/ports.conf
```diff
@@ -43,6 +43,7 @@ devel/phpunit9@php${PRODUCT_PHP}
 devel/pkgconf
 devel/py-Jinja2@py${PRODUCT_PYTHON}
 devel/py-pycodestyle@py${PRODUCT_PYTHON}
+devel/py-pyyaml@py${PRODUCT_PYTHON}
```
fichtner, line 46: what are we fixing here? :)

PR #464, opened 2025-02-24
build/kernel: add dtrace support by default for debug kernel

swhite2:


fichtner:
let's keep this here until 25.1.2 is tagged

fichtner:
@swhite2 I broke this PR due to 54d92acbc8adb0 but it should be easier and cleaner now to achieve this. Can you rebase?

fichtner:
Merged, thanks!

PR #468, opened 2025-03-14
blobless git clone

### LINKED ISSUE #467: Speed up initial clone with --filter=blob:none
On my box, cloning the tools repo and running `make update` takes 45 minutes. (My dev box is terrible and its on a terrible LAN.)

By adding `--filter=blob:none`, I've reduced that to 22 minutes.

Yes, it's twice as fast.

The tradeoff is that some git operations will take longer, as blobs will be downloaded on demand (e.g. doing git bisect or checking out earlier commits).

However, I expect this cost will never be incurred for the majority of workflows. 

For workflows that dive into git history, I expect that the cost of fetching blobs on-demand will still be much less than the cost of fetching everything at clone time.

My understanding of the build system is very superficial, so I am open to having my assumptions corrected.

Ref: https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/

To implement, I think we just need to add the `--filter=blob:none` arg to the `git_clone` function in `build/common.sh`. I attach a PR to demonstrate: #468

fsackur:
fixes #467 

To my superficial understanding, I think this could be merged, but people with a better understanding of the build system should probably correct me if I'm wrong!

### DIFF
--- a/build/common.sh
+++ b/build/common.sh
@@ -344,7 +344,7 @@ git_clone()
 		URL=${PRODUCT_GITBASE}/$(basename ${1})
 	fi
 
-	git clone "${URL}" ${1}
+	git clone --filter=blob:none "${URL}" ${1}
 }
 
 git_pull()


fichtner:
Thanks, I'll test this next week. Could work :)

fichtner:
Sorry for the delay, I am locked in the 7 circles of release engineering these days. Merged, thanks!

PR #475, opened 2025-05-22
Update port configs to support arm64 ignore chipsets

computeronix:
Newer RPIs and other devices, for example, now use arm64 chipsets.  Updated the ports, aux, and other configs to ignore the arm64 chipset during build for the same items that arm chipset is also ignored.

Example build log for an RPI 4

make command -> make base kernel packages arm-3G DEVICE=RPI DEBUG=1 

```
DEVICE=RPI
PRODUCT_ARCH=aarch64
PRODUCT_TARGET=arm64
```

### DIFF
--- a/config/25.1/aux.conf
+++ b/config/25.1/aux.conf
@@ -1,7 +1,7 @@
 #ORIGIN						IGNORE
 devel/binutils
 devel/cmake-core
-lang/go121					arm
-lang/go122					arm
-lang/go123					arm
-lang/rust					arm
+lang/go121					arm,arm64
+lang/go122					arm,arm64
+lang/go123					arm,arm64
+lang/rust					arm,arm64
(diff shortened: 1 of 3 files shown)


fichtner:
I was thinking of removing "arm", but since you suggest it how about just renaming it instead of adding a new one on top?  "arm" was introduced a long time ago and FreeBSD also changed their build infrastructure and the respective hardware doesn't get much better today. This would also entail removing BANANAPI, CLEARFOG and RPI2 device configs?

fichtner:
Looking at this I doubt we need it as we still get exclusions from "aarch64" as per your configuration above. We can always add more, but all seem to be arm64 (TARGET) and aarch64 (ARCH) anyway... I'll take care of this.

computeronix:
Sounds great and from there will request some certain exclusions as I run through a test build to add on top as some certain items need to be excluded from RPI

computeronix:
Also agree with removing RPI2

fichtner:
Ok, if RPI needs build exceptions we have to extend build exceptions to devices since all remaining ARM devices are the same aarch64+arm64. Excluding an arch will exclude for all now. But maybe it all builds. Would be illogical if it didn’t.


Cheers,
Franco

PR #496, opened 2025-10-29
net/miniupnpd: Add build options

Self-Hosting-Group:
This PR adds the required build options `LEASEFILE`/`UPNP_IGDV2`/`UPNP_NOPPP` for the plugin PRs opnsense/plugins#4629 and opnsense/plugins#5005 in tools, and requires the merge of opnsense/ports#244 in ports.


config/25.7/make.conf
```diff
@@ -61,7 +61,7 @@ net_freeradius3_SET=		LDAP MITKRB_PORT MYSQL SQLITE3
 net_freeradius3_UNSET=		HEIMDAL
 net_frr10_SET=			MULTIPATH SNMP
 net_haproxy30_SET=		LUA # PROMEX
-net_miniupnpd_SET=		CHECK_PORTINUSE
+net_miniupnpd_SET=		CHECK_PORTINUSE UPNP_DISABLEPPPCONN UPNP_IGDV2 UPNP_VENDORCFG
```
fichtner: Only `UPNP_IGDV2` in the ports now so we'll park this to avoid errors during build option checks.

One thing of note is that this is usually sorted alphabetically, but we may need to work on the option names anyway. One step at a time :)

Self-Hosting-Group:
Now using `UPNP_NOPPP`, and I removed `UPNP_VENDORCFG`, and unconditionally defined it in the just updated FreeBSD ports PR, and here, we have two commits now.

fichtner (approved):

Issue #378, opened 2023-09-12
Ads security/xray-core to ports and pkgs.


drlebedev:
## Request
Add new package xray-core from Freebsd ports (https://www.freshports.org/security/xray-core/). 

## Requester
Upstream port Maintainer

## Context
The xray core is an encrypted connection proxy that helps to penetrate the censoring firewalls.  Xray is a feature-reach fork of v2ray that is included in the OpnSense package list but is not updated and is not effective today.  Xray is fully config-compatible with v2ray and can be used as a drop-in replacement.  Maintainers of OpnSense pkg repo can consider obsolete v2ray version replacement with xray.

It supports state-of-the-art encryption and obfuscation tools to build custom network topologies and overcome traffic filtering. It supports 

1. Shadowscoks, including Shadowsocks 2022
2. Socks 
3. Vless/VMess
4. Trojan

Traffic can be routed via CDN (Cloudflare) with the support of Websocket and gRPC transport.

## OpnSense pkg repo fit
As Xray is a network security and penetration tool, it fits the core function of OpnSense. OpnSense already supports Shadowsocks (via plugin) and v2ray (obsolete pkg version). Including Xray in packages helps to mitigate security risks from obsolete packages and migrate to the state-of-the-art secure proxy as a drop-in replacement.

drlebedev:
@fichtner Any chance you can take a look on this?

drlebedev:
@fichtner  quick ping on above

fichtner:
@drlebedev Sorry for the delay. Confirmed. v2ray will be replaced by xray-core package in 23.7.8.

Cheers,
Franco

PR #504, opened 2026-03-29
config: add amnezia-kmod, amnezia-tools, amneziawg-go to packages

phprus:
OPNsense currently has a `xray-core` (#378) package for traffic obfuscation, but without a plugin for the web interface.

I propose to add support for the Amnezia VPN ( https://docs.amnezia.org/documentation/amnezia-wg/ ) protocol in OPNsense.

> Amnezia VPN is a VPN protocol that is backward compatible with the WireGuard VPN protocol. It offers protection against detection by Deep Packet Inspection (DPI) systems. At the same time, it retains the simplified architecture and high performance
>
> The Amnezia VPN protocol has issues with detection due to distinctive packet signatures. Amnezia addresses this problem by employing advanced obfuscation methods, allowing its traffic to blend seamlessly with regular internet traffic.

There are two open source plugins for managing Amnezia connections for OPNsense:

- https://github.com/MrTheory/os-amneziawg ( https://forum.opnsense.org/index.php?topic=51023.0 )
- https://github.com/antspopov/opnsense_amnezia_plugin

Adding `net/amnezia-kmod`, `net/amnezia-tools`, `net/amneziawg-go` packages to the OPNsense repository will make it easier to use these plugins.

`amnezia-kmod` - kernel module for Amnezia protocol (version 1.0 and 1.5, and partial support of 2.0).
`amnezia-tools` - userspace tooling for using and configuring Amnezia tunnels.
`amneziawg-go` - userspace protocol implementation from Amnezia team (https://github.com/amnezia-vpn/amneziawg-go/).

@fichtner, Please review this PR, thank you.

### DIFF
--- a/config/26.1/ports.conf
+++ b/config/26.1/ports.conf
@@ -107,6 +107,9 @@ net-mgmt/zabbix72-proxy
 net-mgmt/zabbix74-agent
 net-mgmt/zabbix74-proxy
 net/addrwatch
+net/amnezia-kmod
+net/amnezia-tools
+net/amneziawg-go
 net/bird2
 net/chrony
 net/dpinger


fichtner:
At first glance as already mentioned in the forum we don’t need another one alongside netbird and tailscale. Someone asked to build a plugin for xray-core lately too.

I’m also not enthusiastic about another kernel level solution.


Cheers,
Franco

phprus:
Thank you for your reply!

Tailscale and NetBird use WireGuard, which is successfully blocked by DPI.
Xray and Amnezia are not blocked.

Xray is already included in OPNsense, but Amnezia, unfortunately, is not included.

fichtner:
I see the benefit of it and we’ve been welcome to such things in the past such as v2ray/xray, openvpn xor, softether, etc. All of which had their problems from inclusion and adoption and long term maintenance.

WireGuard made the scene much more complicated for mostly the right reasons, but now all descendent projects with the best intentions need to drag a behemoth through each OS and distribution stack which makes plugins work even more complicated than it ever was and at the same time being a commercial driver for whoever started such a project.

Perhaps we should make a survey in the forum about inclusion and tie some conditions to it, but I’m also not very interested in reviewing another large PR for a new VPN plugin. The biggest issue is that we can’t take things back easily after realising this wasn’t the best thing to do.


Cheers,
Franco

phprus:
> Perhaps we should make a survey in the forum about inclusion and tie some conditions to it, but I’m also not very interested in reviewing another large PR for a new VPN plugin.

New plugin or new packages from ports?

Plugins for Amnezia, xray, etc may be in separate repositories outside of opnsense/plugins.
Building a plugin is a simple, unlike building packages from ports.

I would suggest only adding packages from FreeBSD ports.


> The biggest issue is that we can’t take things back easily after realising this wasn’t the best thing to do.

Before creating this PR, I spent several days attempting to automate the build of the above ports using poudriere, but I encountered difficulties creating jails.

The archives
https://pkg.opnsense.org/FreeBSD:14:amd64/26.1/sets/
base-26.1.3-amd64.txz
kernel-26.1.3-amd64.txz

and the manually created src.txz seem insufficient, as they lack the opnsense code itself and its repository settings. Poudriere also rebuilds dependencies that are already exists in the OPNSense repository.

Furthermore, even the existing archives don't contain the latest version, which can complicate the build.



Having a jail archive for poudriere for current OPNsense versions would make it easier to build additional ports independently.

Then this (and similar) PRs would be unnecessary.

phprus:
Is there a known way to create a full-featured jail for poudriere?

The instructions at https://forum.opnsense.org/index.php?topic=21739.0 use archives that do not contain OPNsense repository settings.

fichtner:
> Building a plugin is a simple, unlike building packages from ports.

I think the reverse is true. You’re also asking to provide this to the OPNsense packages for the foreseeable future, including complications and expertise getting it to compile, distributing through mirror maintainers. I just think it’s a lot to ask.

Mind you this is not a 10000 lines of code no dependency plugin that builds in 10 seconds and distributes as a 2 kb package file.


Cheers,
Franco

phprus:
I apologize for my bad English.

> I think the reverse is true. You’re also asking to provide this to the OPNsense packages for the foreseeable future, including complications and expertise getting it to compile, distributing through mirror maintainers. I just think it’s a lot to ask.

Building the plugin, not writing it code. Building the plugin package is simple:

```
opnsense-code tools plugins
cp my_plugin /usr/plugins/net
cd /usr/plugins/net/my_plugin
make package
```

But building FreeBSD port with dependencies into packages is not simple process.


I am not asking you to distribute plugins through the OPNsense infrastructure. I don't ask them to review or add to `opnsense/plugins`.
I only asked for packages that are already in FreeBSD ports.

This can also be difficult for OPNsense, and that's why I'm asking for help in creating an environment in which third-party developers can do all this themselves. Help in creating poudriere jail for OPNsense.

Monviech:
It would be better to wrap wireguard inside something that is independant from it, like wstunnel.

https://github.com/erebe/wstunnel

Everything that alters and ships wireguard directly is very inflexible and technical debt once DPIs got a hang of it too. (static targets are prone to fingerprinting)

The more widespread and mainstream such an obfuscation technique becomes the more likely it will be blocked as well in time.

Better to be able to change the technique independantly from the tunneling protocol in the long run.

Also, wstunnel seems sponsored by an NL company.

AdSchellevis:
`wstunnel` might actually be a practical addition for cases like these, we do have `stunnel` already, but the wstunnel project looks rather vivid (and shouldn't be very difficult to wrap in a plugin)

phprus:
Wstunnel is not a VPN protocol, is a proxy.
Amnezia VPN is a VPN protocol that allows to create private networks.

Monviech:
wstunnel is a dedicated obfuscation tool with multiple ways to mask your traffic. It can wrap around any vpn protocol. Your point is moot.

Amneziawg also needs both sides supporting it, just like any proxy and obfuscation tool.

What you probably wanted to say is "I want the simplest most streamlined setup with the least work for me in the future."

Who is expected to carry the long-term cost? Decoupling layers has always been better in the long term.

Monviech:
Additionally I would give this a read:

https://reviews.freebsd.org/D51239?id=158287

Jason (the author of WireGuard) essentially says the same thing, transport and obfuscation belongs on a different, faster moving layer than the kernel.

Essentially all the proxies that exist are the correct layer for this, and welding something onto a vpn protocol has not the velocity or flexibility to avoid DPI detection long term.

phprus:
`amneziawg-go` is a user-space implementation.
`amnezia-kmod` is a kernel module that can be updated more frequently than the kernel.

wstunnel is a great solution, but it uses TCP.
Tunneling TCP traffic over a TCP-based tunnel can reduce throughput by tens of times or more.

Tragen:
> wstunnel is a great solution, but it uses TCP. Tunneling TCP traffic over a TCP-based tunnel can reduce throughput by tens of times or more.

I haven't used wstunnel yet but the GitHub page shows how to use UDP.

AdSchellevis:
`wstunnel` might be something to consider for people needing to fix this particular issue, Amnezia for a lot of reasons (see also  https://forum.opnsense.org/index.php?topic=48357.msg265079#msg265079) simply isn't.  Our tracker isn't intended for marketing purposes.

fichtner:
Closed as previously mentioned.

